PR #5, opened 2018-05-07
Cake 3141 - track user actions 

nmonterroso:
@Wikia/cake 

Tracks user actions when using the modal. Whether or not including the beacon is considered personably identifiable information is TBD. If it is the tracking implementation might change, but its usage and interface should remain the same.


JoshuaRogan:
Can we add `track` to the default options? Otherwise LGTM

nmonterroso:
Ah, good catch. It was initially there but must have gotten lost in the merge. ee8f2b4

PR #6, opened 2018-05-07
update entry point to class

nmonterroso:
@Wikia/cake 

Updates the main entry point to use a class and returns an instance of that class when instantiated. This will make [-3132](https://wikia-inc.atlassian.net/browse/CAKE-3132) easier since the app can just call `.reset()` rather than having to know the name of the cookie externally and re-initializing possibly without knowing the passed `options`. This might also make testing easier.

I also moved all the hot reloading logic to `index-dev` which isn't used when making the production build. This is removed by webpack but initially I wanted to also expose all functionality as es5 modules in addition to making the entry point a class, but I ran into an issue where `styles.scss` was getting ignored in the babel transpilation, similar to what @drsnyder is seeing with unit tests. Hopefully the fix is the same, and once that's figured out I'll re-add the es5 modules as part of the build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,7 @@
+{
+  "presets": ["env", "preact"],
+  "plugins": [
+    "transform-class-properties",
+    "transform-object-rest-spread"
+  ]
+}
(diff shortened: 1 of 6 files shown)


JoshuaRogan:
```
/Users/joshrogan/projects/tracking-opt-in/webpack.config.js:111
    ...topLevelOptions,
```

PR #7, opened 2018-05-08
CAKE-3137 add the scaffolding for class and component testing

drsnyder:
https://wikia-inc.atlassian.net/browse/CAKE-3137

Adds Karma + Mocha + Chai for testing. 

In the first pass at this, I tried using Enzyme for the rendering test but ran into some issues with the helper libraries provided by the community. Either they wouldn't run due to internal errors or `wrapper.find(<classname>)` would only work at the top-most level.

I decided to backtrack and use the simplest approach which is to render the component directly. This may have limitations of it's own down the road if we need to invoke lifecycle methods. We can revisit if necessary.

I added two CLI options for testing:
 - `yarn test` runs a single pass
 - `yarn test:watch` rebuilds, and re-runs the tests

The testing is nowhere near complete. I wanted to go ahead and get this merged so we can start backfilling the tests and add tests as we go.

@Wikia/cake


karma.conf.js
```diff
@@ -0,0 +1,118 @@
...
+              rules: [
+                  {
+                      test: /\.js$/,
+                      include: path.resolve(__dirname, './src'),
+                      use: [{
+                          loader: 'babel-loader',
+                          options: {
+                              presets: ['env', 'preact'],
+                              plugins: [
+                                  'transform-class-properties',
```
JoshuaRogan: Can this use the `.babelrc` file instead?

drsnyder: Ah. Yes a6bae4f.

karma.conf.js
```diff
@@ -0,0 +1,112 @@
...
+                  },
+                  {
+                      test: /\.s?css$/,
+                      use: [
+                          {
+                              loader: 'style-loader',
+                              options: {
+                                  sourceMap: true,
+                              }
+                          },
```
drsnyder: This block addressed the issue I was having with the `style` object in `App` and `App_test`.

drsnyder:
I looked at consolidating some of the babel settings. I agree it's lame to copy them but it's also hard to follow when part of the config is included (as we're doing in the FC). For my part, I find it easier to look at them as a contiguous block.

drsnyder:
Added code coverage in 64a967d.

drsnyder:
I changed the `mode` to `'development'`. I added a `debugger` statement to `render` and the code was transpiled but readable.

PR #8, opened 2018-05-09
 CAKE-3135 | NPM Testing and Designs

JoshuaRogan:
* Added test npm package
* Designs except background
  * ~The svg provided was 165KB so I asked if JJ could go back and come up with something smaller and simpler.~
  * JJ is okay with removing it

@Wikia/cake

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tracking-opt-in-test",
-  "version": "1.0.4",
+  "version": "1.0.5",
   "main": "dist/tracking-opt-in.js",
   "license": "MIT",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tracking-opt-in-test",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "main": "dist/tracking-opt-in.js",
   "license": "MIT",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "tracking-opt-in-test",
-  "version": "1.0.6",
+  "name": "tracking-opt-in",
+  "version": "1.0.7",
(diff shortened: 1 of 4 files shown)


src/ContentManager.js
```diff
@@ -1,62 +1,87 @@
...
         initialHeadline: 'This site uses cookies',
+        initialBodyText: 'FANDOM and our partners use technology such as cookies on our site to personalize content and ads, provide social media features, and analyze our traffic. Can we use this technology to deliver you a personalized experience?',
         secondHeadline: 'Are you sure?',
-        buttonAccept: 'Accept all cookies',
-        buttonReject: 'Reject all cookies',
+        secondBodyText: 'By selecting no, you may see ads that are less relevant to you and certain features on the site may not work as intended. You can update your preferences at any time by visiting our Privacy Policy.',
+        buttonAccept: 'YES',
+        buttonReject: 'NO',
+        secondButtonAccept: 'Accept All Cookies',
+        secondButtonReject: 'Reject All Cookies',
```
vforge: AFAIK it shouldn't say "all cookies"

JoshuaRogan: The final copy is still in flux but this is what is provided in the mocks. I just shared you on that.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "tracking-opt-in",
-  "version": "1.0.0",
+  "name": "tracking-opt-in-test",
```
vforge: I'm curious, what's the reason for rename? It was matching repo's name, `tracking-opt-in` before

src/ContentManager.js
vforge: Thanks! I confirmed that we'll change copy later; thanks!

package.json
JoshuaRogan: I'm leaving it as test until the oficial wikia account is created. It is currently published under my name.

PR #9, opened 2018-05-09
Styles in script

vforge:
- always put styles in the script (clean-up the webpack config BTW)
- `dist` version does not need anything (`js-cookie` and `preact` are, in fact, dev dependencies)
- remove obsolete `mini-css-extract-plugin` package

/cc @Wikia/cake

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,16 +15,17 @@
     "copy-webpack-plugin": "^4.5.1",
     "cross-env": "^5.1.4",
     "css-loader": "^0.28.11",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -33,8 +34,6 @@
     "webpack-serve": "^0.3.2"
   },
   "dependencies": {
```
JoshuaRogan: Ahh this is why they were added into our repo. Good catch.

PR #11, opened 2018-05-10
License, Privacy Policy Links, and Updated Copy

JoshuaRogan:
@Wikia/cake


src/components/App.js
```diff
@@ -96,15 +59,19 @@ class App extends Component {
...
+                        <div>
+                            {content.bodyParagraphs.map((paragraph) =>
+                                <p>{paragraph}</p>
+                            )}
+                        </div>
                     </div>
                     <div className={styles.buttons}>
-                        <div className={styles.buttonPrimary} onClick={onAccept}>
-                            {acceptText}
+                        <div className={styles.buttonPrimary} onClick={this.onInitialAccept}>
```
nmonterroso: minor: can we consolidate `onInitialAccept/accept` into a single function? I don't think there's a reason to have them separated now. Same for rejections.

JoshuaRogan: Done 👍

PR #17, opened 2018-05-11
Remove the manual checkout from the jenkins pipeline script

drsnyder:
After taking a closer look at the jenkins UI (it's a maze) I noticed there were additional options that could be provided to the initial SCM checkout. In those options you can specify that you want the repo to be cleaned before and after the checkout. That should ensure a pristine repo so we can remove the redundant internal checkout.

I must have been so excited to get the deploy working in the FC that I overlooked this option. If it works as expected here we can update that workflow as well.

As part of this, I also added tooling to validate the jenkins file. Having to commit, push, go to jenkins, rebuild, repeat slows down the iteration on building the release pipeline. This should help improve that process by giving the developer a local means for validating the file without having to go through the steps above. To validate, run `yarn validate-jenkins`.

@Wikia/cake

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "prepublishOnly": "yarn build",
     "test": "karma start",
     "test:watch": "karma start --auto-watch --no-single-run",
-    "test:selenium": "wdio webdriver.conf.js"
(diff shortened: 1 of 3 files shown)


nmonterroso:
LGTM

PR #18, opened 2018-05-11
Allow some constants to be imported without relying on dependencies

JoshuaRogan:
@vforge Let me know if you'd like to have any other constants here. 

@Wikia/cake

### DIFF
--- a/src/OptInManager.js
+++ b/src/OptInManager.js
@@ -1,5 +1,5 @@
 import Cookies from 'js-cookie';
-import { DEFAULT_COOKIE_NAME, COOKIE_STATUS } from 'src/constants';
+import { DEFAULT_COOKIE_NAME, COOKIE_STATUS } from './constants';
 
 const DEFAULT_ACCEPT_COOKIE_EXPIRATION = 18250; // 50 years in days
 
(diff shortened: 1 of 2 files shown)


nmonterroso:
I'm curious - why do you need these constants externally? IMO you shouldn't really be concerned with the name of the cookies and their values, but with whether or not the user has accepted tracking. We can add a function to expose that if you'd like.

vforge:
@nmonterroso We need it when testing for the tracking - see https://github.com/Wikia/gdpr-tracking-tester/blob/9726626dd024d7c53822177830916264e9a4037f/src/utils/fetchLogsFromUrls.js#L29-L30

The function won't do because we're not running any code on the page - we're just opening the page with a cookie (or not).

vforge:
Other solution would be to hard-code cookie name and values along with the URL as the cookie name can differ from product to product (in theory).

nmonterroso:
The name of the cookie name _should_ be consistent across products since otherwise the user will be prompted with a modal when visiting different fandom apps.

My vote would be for usage like this to be hard-coded on the test side since otherwise the dependency now includes the structure and constant names of the library rather than on the functionality.

FWIW, we have our own [selenium tests](https://github.com/Wikia/tracking-opt-in/blob/master/selenium/test.js) that should work on any site. Since the app url is configureable this can be a jenkins task that takes an app url and verifies functionality in a centralized place. Otherwise if/when the implementation changes anyone who has their own integration tests will have to update them.

vforge:
@nmonterroso @JoshuaRogan As the cookie name/value are for testing purposes only, let's abandon this idea to get those values from the source of the library.

PR #20, opened 2018-05-11
Cake 3158 - start test container with latest build during release process

nmonterroso:
@Wikia/cake 
https://wikia-inc.atlassian.net/browse/CAKE-3158

Creates a docker image and deploys that image to k8s during the release process, and then deletes those resources from k8s when the process finishes, regardless of whether or not the release completed successfully or not.

### DIFF
--- a/release/demo-container/k8s.hbs
+++ b/release/demo-container/k8s.hbs
@@ -53,6 +53,8 @@ kind: Ingress
 metadata:
   name: {{serviceName}}
   namespace: {{namespace}}
+  labels:
+    app: {{serviceName}}
 spec:
   rules:
   - host: {{appHost}}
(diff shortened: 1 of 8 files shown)


release/demo-container/k8s.hbs
```diff
@@ -0,0 +1,65 @@
...
+      - name: {{serviceName}}
+        image: {{image}}
+        livenessProbe:
+          httpGet:
+            path: /
+            port: 80
+          initialDelaySeconds: 30
+          timeoutSeconds: 3
+          periodSeconds: 20
+        resources:
```
JoshuaRogan: Do we have any documentation on choosing defaults for the number of resources? It seems like a lot for something this small.

nmonterroso: We don't really have any guidelines. I copied the config from fandom-creator and changed it to be a handlebars template. I'll tune this down though, since it is a lot.

nmonterroso: ac53c43

PR #21, opened 2018-05-14
CAKE-3150 | Bump version, releast to npm, and push tag during release

JoshuaRogan:
There were a few snakes while trying to figure this out. `yarn publish` doesn't behave exactly as `npm publish`. It appears there isn't a way to publish with yarn without changing the version number -- which is there is no way to do automatically. 

https://github.com/yarnpkg/yarn/issues/610
https://github.com/yarnpkg/yarn/commit/0f713c1e4c561fbcfdedba3e6c83afdd3166d301


release/jenkins
```diff
@@ -62,11 +62,17 @@ pipeline {
     stage('release') {
       when {
         expression {
-          return params.pipeline == 'release'
+          return params.pipeline == 'release' && params.branch == 'master'
         }
       }
       steps {
-        sh "echo '...creating tag and releasing to npm...'"
+        sshagent(credentials: ['bd3cf712-39a7-4b16-979e-ff86208ab2ea']) {
```
nmonterroso: Can this be an env var passed in from jenkins? We're trying to keep all of our env-specific configs outside of the repo.

JoshuaRogan: Updated

PR #27, opened 2018-05-15
CAKE-3168 | Add query param override

JoshuaRogan:
Adds the ability to override the cookie check with a query param `tracking-opt-in-accepted (default). If the query param is in `window.location.search` it will set the cookie for you. 

@Wikia/cake

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ The following options are accepted:
 - `countriesRequiringPrompt` - array of country codes that require tracking opt-in. See [`GeoManager`](https://github.com/Wikia/tracking-opt-in/blob/master/src/GeoManager.js) for the defaults.
 - `language` - Override the language used to display the dialog text. Defaults to `window.navigator.language` if available, otherwise to `en`.
 - `preventScrollOn` - Prevent scrolling on the specified element when the dialog is shown. Can be either an element or query selector passed to `document.querySelector`. Defaults to `'body'`, set to `null` to prevent this behavior.
+- `queryParam` - The name of the query param to forcefully set the accepted status. Defaults to `tracking-opt-in-accepted`
 - `track` - whether to track impressions and user consent/rejections. Defaults to `true`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -38,6 +38,7 @@ The following options are accepted:
 - `countriesRequiringPrompt` - array of country codes that require tracking opt-in. See [`GeoManager`](https://github.com/Wikia/tracking-opt-in/blob/master/src/GeoManager.js) for the defaults.
 - `language` - Override the language used to display the dialog text. Defaults to `window.navigator.language` if available, otherwise to `en`.
 - `preventScrollOn` - Prevent scrolling on the specified element when the dialog is shown. Can be either an element or query selector passed to `document.querySelector`. Defaults to `'body'`, set to `null` to prevent this behavior.
+- `queryParam` - The name of the query param to forcefully set the accepted status. Defaults to `tracking-opt-in-accepted`
```
nmonterroso: I think this should be `force-tracking-opt-in-status` and accept something truthy/falsey, since the user might also reject cookies, in which case we also don't want to reprompt them.

src/OptInManager.js
```diff
@@ -17,13 +18,20 @@ function getCookieDomain() {
...
-        this.domain = domain || getCookieDomain();
+        this.domain = getCookieDomain();
+        this.queryParam = queryParam || DEFAULT_QUERY_PARAM_NAME;
     }
 
     getValue() {
+        const cookie = Cookies.get(this.cookieName);
+
+        if (cookie === undefined && window.location.search.indexOf(this.queryParam) !== -1) {
+            this.setTrackingAccepted();
```
nmonterroso: I think it's a little strange that `getValue` is conditionally setting a cookie. I think this should be a separate function like `setForcedStatusFromQueryParams` (or something similar) and called from `main`. That way there are no side effects on a getter and the flow is more explicit.

JoshuaRogan:
@nmonterroso I made some updates but still kept the logic within the `OptInManager` so that is the only place that modifies the tracking cookie.

PR #34, opened 2018-05-17
Fix typo in polish translation

fraszczakszymon:
I'm not sure where is the source of translations. Should I fix it somewhere else?


nmonterroso:
@fraszczakszymon we copied our translations from the spreadsheet [here](https://docs.google.com/spreadsheets/d/1KkUwvcmYcZNx4B4hT6bibRdpoaSV5JrmIB28RQ7Jbl0/edit#gid=728056174)

PR #35, opened 2018-05-17
CAKE-3159 | Browser stack tests

JoshuaRogan:
* Run browserstack tests on jenkins
* Generate allure and junit reports 

TODO: 
- [x] Enable different domain testing


release/jenkins
```diff
@@ -56,7 +56,7 @@ pipeline {
 
     stage('BrowserStack') {
       steps {
-        sh "echo '...running the browserstack tests...'"
+        sh "BUILD_ID=${currentCommit} TEST_URL=${env.DOMAIN ?: "http://gdpr-modal-test-${currentCommit}.wikia-dev.us"}« yarn test:selenium"
```
nmonterroso: In the past I've tried to keep this file as wikia-free as possible, since it's open source. 

I added the following env vars for this purpose to the build:
```
APP_HOST_PREFIX=gdpr-modal-test-
APP_HOST_SUFFIX=.wikia-dev.us
```
What do you think about using those?

nmonterroso: Also what's up with the `«` before `yarn test:selenium`?

release/jenkins
```diff
@@ -84,29 +84,40 @@ pipeline {
         }
       }
     }
+
+    stage('reports') {
+      steps {
+        script {
+          allure([includeProperties: false, jdk: '', properties: [], reportBuildPolicy: 'ALWAYS', results: [[path: 'reports/selenium/allure'], [path: 'reports/karma/allure']]])
```
nmonterroso: Any reason not to have this in `always`? I think that if this is in a `stage` then if the tests fail the reports won't get generated. Am I wrong in that?

selenium/test.js
```diff
@@ -3,6 +3,7 @@ const assert = require('assert');
 const url = process.env.TEST_URL || 'http://localhost:3000';
 const countryRequiringConsent = process.env.COUNTRY_REQUIRING_CONSENT || 'PL';
 const countryNotRequiringConsent = process.env.COUNTRY_NOT_REQUIRING_CONSENT || 'CA';
+const domain = url.match('^(?:https?:\/\/)?(?:[^@\/\n]+@)?(?:www\.)?([^:\/\n]+)')[1];
```
nmonterroso: Can you add a note on what this is supposed to be doing?

nmonterroso:
LGTM

PR #36, opened 2018-05-18
Update Translations

### LINKED ISSUE #29: Split non-English messages in modal's body into three paragraphs
Wasn't sure where should I report it, so I decided to create an issue on GitHub. There's a slight aesthetical issue with non-English messages being not formatted in the same way as English strings are. Since these modal will be shown to really many users, I believe fixing this could be worth it.

**English:**
en

**German:**
de

There seems to be difference between English and non-English texts in the implementation. For example German lang has whole text as one string: https://github.com/Wikia/tracking-opt-in/blob/9983da1694e56d6a13e37ba7732904d8c96768a2/src/ContentManager.js#L5

...but on the other hand for English lang has same text is split to three strings: https://github.com/Wikia/tracking-opt-in/blob/9983da1694e56d6a13e37ba7732904d8c96768a2/src/ContentManager.js#L17
https://github.com/Wikia/tracking-opt-in/blob/9983da1694e56d6a13e37ba7732904d8c96768a2/src/ContentManager.js#L18
https://github.com/Wikia/tracking-opt-in/blob/9983da1694e56d6a13e37ba7732904d8c96768a2/src/ContentManager.js#L19

I wanted to submit Pull Request with non-English texts splitted, although I don't feel competent enough to recognize these sentences in other languages. ;) I guess there might be needed some help from LOC team.

JoshuaRogan:
Break translations into proper paragraphs provided from Keiko

@Wikia/cake 

resolves #29

### DIFF
--- a/src/ContentManager.js
+++ b/src/ContentManager.js
@@ -2,7 +2,9 @@ export const langToContent = {
     de: {
         headline: 'Diese Website verwendet Cookies',
         bodyParagraphs: [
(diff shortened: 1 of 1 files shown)


nmonterroso:
👍

PR #37, opened 2018-05-21
CAKE-3179 | Add external test

JoshuaRogan:
Runs just the selenium tests on an external URL. 

I spent some time trying to create a shared lib (https://jenkins.io/doc/book/pipeline/shared-libraries/#defining-global-variables) but couldn't get the imports working properly and it was quite time consuming to test. It turns out there isn't too much commonality anyway. 

Remote Trigger: `http://jenkins:8080/view/CAKE/view/tracking-opt-in/job/external%20test/buildWithParameters?token=tracking-opt-in-external-test&testUrl=<url>&slackChannel=cake-team-tech`

@Wikia/cake

### DIFF
--- a/selenium/jenkins
+++ b/selenium/jenkins
@@ -25,31 +25,25 @@ pipeline {
 
     post {
         success {
-            when {
-                expression {
-                    return params.slackChannel != ''
-                }
+            if  (params.slackChannel != '') {
+                slackSend(
+                        channel: "#${params.slackChannel}",
+                        color: 'good',
+                        failOnError: true,
+                        message: ":checkmark: tracking-opt-in external test passed <${env.BUILD_URL}|succeeded at ${params.testUrl}>"
+                )
             }
-            slackSend (
-                    channel: "#${params.slackChannel}",
-                    color: 'good',
-                    failOnError: true,
-                    message: ":checkmark: tracking-opt-in external test passed <${env.BUILD_URL}|succeeded at ${params.testUrl}>"
-            )
         }
 
         failure {
-            when {
-                expression {
-                    return params.slackChannel != ''
-                }
+            if  (params.slackChannel != '') {
+                slackSend(
+                        channel: "#${params.slackChannel}",
+                        color: 'warning',
+                        failOnError: true,
+                        message: ":warning: tracking-opt-in external test failed <${env.BUILD_URL}|failed> at ${params.testUrl}"
+                )
             }
-            slackSend (
-                    channel: "#${params.slackChannel}",
-                    color: 'warning',
-                    failOnError: true,
-                    message: ":warning: tracking-opt-in external test failed <${env.BUILD_URL}|failed> at ${params.testUrl}"
-            )
         }
 
         always {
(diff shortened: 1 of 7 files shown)


selenium/jenkins
```diff
@@ -0,0 +1,53 @@
...
+    stages {
+        stage('dependencies') {
+            steps {
+                sh "yarn install --frozen-lockfile"
+            }
+        }
+
+        stage('BrowserStack') {
+            steps {
+                sh "BUILD_ID=${params.testUrl}-${currentCommit} TEST_URL=${params.testUrl} yarn test:selenium"
```
nmonterroso: We probably want to overwrite the project that's used in `webdriver.conf.js`

selenium/jenkins
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+    }
+
+    post {
+        success {
+            slackSend (
+                    channel: '#cake-team-tech',
```
nmonterroso: Not sure we want our channel to get spammed with test results for other teams. What about a parameter that when absent doesn't send the message?

JoshuaRogan:
Updated the trigger to `http://jenkins:8080/view/CAKE/view/tracking-opt-in/job/external%20test/buildWithParameters?token=tracking-opt-in-external-test&testUrl=<url>&slackChannel=cake-team-tech` and appended the url to the project name. I'm not entirely sure how the 'project' name is used though.

nmonterroso:
The project name is used by [`webdriver.conf`](https://github.com/Wikia/tracking-opt-in/blob/master/webdriver.conf.js#L73). The badge on the README is tied to a specific project, so if we have the external tests report to a different project then a failing external test will not mark the core project as failed.

JoshuaRogan:
Do you think appending the url is sufficient? It should make it so any team can then have a badge

nmonterroso:
> Do you think appending the url is sufficient? It should make it so any team can then have a badge

I wonder if this should be another parameter. `projectName` or something. That way the `project` in BrowserStack can be specified by the teams. What do you think about that?

selenium/jenkins
```diff
@@ -0,0 +1,54 @@
...
+        stage('dependencies') {
+            steps {
+                sh "yarn install --frozen-lockfile"
+            }
+        }
+
+        stage('BrowserStack') {
+            steps {
+
+                sh "BUILD_ID=${params.testUrl}-${currentCommit} TEST_URL=${params.testUrl} PROJECT_NAME=${params.projectName} ${params.testIE11 ? "TEST_IE_11=true" : ""} yarn test:selenium"
```
nmonterroso: What do you think about using `params.projectName` instead of `params.testUrl`? Seems like the name would be nicer to look at.

selenium/test.js
```diff
@@ -74,7 +77,7 @@ describe("BrowserStack: ", () => {
             assert.equal(cookie.value, cookieState.accepted);
             ensureNoPrompt();
         });
-
+        //
```
nmonterroso: leftover?

PR #38, opened 2018-05-23
CAKE-3194 Change simplified to zh and  add zh-TW and zh-HK using traditional

drsnyder:
Chrome uses the following language tags for Chinese:

- Chinese: `zh`
- Chinese (Simplified): `zh-CN`
- Chinese (Hong Kong): `zh-HK`
- Chinese (Traditional): `zh-TW`

Changed simplified from `zh-hans` to `zh`. This should be a reasonable default since there are more simplified readers than traditional. The language lookup logic falls back to the first 2 characters so `zh-CN` will correctly land on `zh`.

Changed traditional from `zh-hant` to `zh-TW` and copied `zh-TW` to `zh-HK`. Traditional is used in Hong Kong so this is a more appropriate setting for those users.

@Wikia/cake

### DIFF
--- a/src/ContentManager.js
+++ b/src/ContentManager.js
@@ -133,7 +133,7 @@ export const langToContent = {
         partnerLinkText: 'Список партнеров',
         partnerLink: 'http://fandom.wikia.com/partner-list',
     },
-    'zh-hans': {
+    'zh': {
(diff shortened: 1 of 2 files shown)


nmonterroso:
LGTM

PR #42, opened 2018-05-24
CAKE-3172 | Whitelist prod privacy and partner pages

JoshuaRogan:
Whitelists the `privacyLink` and `partnerLink` from the content manager. Not sure how to write a selenium test for this until it's actually live on those pages. 

@Wikia/cake

### DIFF
--- a/src/TrackingOptIn.js
+++ b/src/TrackingOptIn.js
@@ -1,13 +1,15 @@
 import {h, render} from "preact/dist/preact";
 import App from "./components/App";
+import {parseUrl} from "./urlUtils";
 
 class TrackingOptIn {
-    constructor(tracker, optInManager, geoManager, contentManager, options) {
+    constructor(tracker, optInManager, geoManager, contentManager, options, location) {
         this.tracker = tracker;
         this.optInManager = optInManager;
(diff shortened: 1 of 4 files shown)


src/TrackingOptIn.js
```diff
@@ -1,13 +1,15 @@
 import {h, render} from "preact/dist/preact";
 import App from "./components/App";
+import {parseUrl} from "./urlUtils";
 
 class TrackingOptIn {
-    constructor(tracker, optInManager, geoManager, contentManager, options) {
+    constructor(tracker, optInManager, geoManager, contentManager, options, location) {
```
JoshuaRogan: Not the most elegant solution but required to test. I also needed to keep the functionality in `TrackingOptIn` to get access to the contentManager.

nmonterroso: I think it's great to have this parameter _because_ it makes the code more testable. I'd actually say we should take it one step further and make the param required, and pass in `window.location` from `index`

JoshuaRogan:
Just updated with an additional test for reset and moved window.location up a level.

nmonterroso:
LGTM

PR #48, opened 2018-07-09
Timeout bump

JoshuaRogan:
Increase timeouts to prevent erroring out. 

@Wikia/cake

### DIFF
--- a/webdriver.conf.js
+++ b/webdriver.conf.js
@@ -173,7 +173,7 @@ exports.config = {
     coloredLogs: true,
     reporters: ['junit', 'concise', 'allure'],
     mochaOpts: {
-        timeout: 300000,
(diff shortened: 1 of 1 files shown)


nmonterroso:
lgtm

webdriver.conf.js
```diff
@@ -173,7 +173,7 @@ exports.config = {
     coloredLogs: true,
     reporters: ['junit', 'concise', 'allure'],
     mochaOpts: {
-        timeout: 300000,
+        timeout: 400000,
```
hakubo: I wonder what happens in these 6.6 min?

JoshuaRogan: It's the entire suite of browserstack tests across several devices.

PR #49, opened 2018-07-10
ADEN-7303 Adds Consent Management Provider

leshniak:
**Ticket:** https://wikia-inc.atlassian.net/browse/ADEN-7303
**Docs:** https://wikia-inc.atlassian.net/wiki/spaces/GDPR/pages/385220674/CMP

Due to compatibility issues with our partners, we want to implement IAB CMP framework instead of our custom solution, which is not supported by anyone but us.
Currently, the CMP module exists in the Oasis codebase, but tracking-opt-in modal seems to be a better place for it.

To do:
- [x] fix and create unit tests for CMP
- [ ] ~~add logs~~ not implemented due to lack of log system


src/ContentManagementProvider-test.js
```diff
@@ -0,0 +1,6 @@
+import {assert} from 'chai';
+import ConsentManagementProvider from './ConsentManagementProvider';
+
+describe('ConsentManagementProvider', () => {
+
```
JoshuaRogan: 😢

package.json
```diff
@@ -9,16 +9,20 @@
...
     "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-decorators-legacy": "^1.3.5",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-preact": "^1.1.0",
     "browserstack-local": "^1.3.1",
     "chai": "^4.1.2",
+    "consent-string": "^1.2.4",
+    "core-decorators": "^0.20.0",
```
JoshuaRogan: I'm a little hesitant to add decorators to this lib. As nice as they are, the proposal isn't yet finished and doesn't look to be finished anytime soon. It may cause confusion down the road if anything changes to the spec. If there is a strong use case for it I can be convinced.

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,388 @@
...
+                            callId: call.callId
+                        }
+                    };
+                    event.source.postMessage(returnMsg, '*');
+                });
+            }
+        }, false);
+    }
+
+    get vendorConsentCookie() {
```
nmonterroso: In general we've been avoiding these sorts of getters/setters and making explicit `getX` and `setX` functions. They allow us to more easily extend functionality when needed, such as `getVendorConsentCookie(valueIfNotSet=someSaneDefault)` and `setVendorConsentCookie(consentString, expirationTime)`. 

If we wanted to make that sort of change here we'd have to first convert these functions over to normal functions before extending them.

It's also a bit confusing since when reading usages of these functions I'd expect property setting/getting to set/get something within the class, not something globally in the browser.

src/components/App.js
```diff
@@ -51,17 +56,27 @@ class App extends Component {
...
+            allowedPublisherPurposes
+        ));
         this.props.onRequestAppRemove();
-        this.props.options.onAcceptTracking();
     };
 
     reject = () => {
         this.track(ACTION_CLICK, 'reject-screen-1');
-        this.props.optInManager.setTrackingRejected();
+        this.props.onUserConsent([], [], []);
```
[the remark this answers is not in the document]
leshniak: uhm, my mistake, it should be `this.props.onUserConsent(userConsent([], [], []));`. Those are  arrays of vendor/purposes IDs.
In case of CMP, there is no distinction between "global" accept and reject actions. From the API perspective there is only a list of consents for particular vendors/purposes. I think we should follow this pattern, even if our consent logic looks different at the top. Many CMPs allow to customize the consents, we could iterate through vendors and purposes and show checkboxes for them, like in [Quantcast Choice](https://www.quantcast.com/gdpr/consent-management-solution/).

See this spec https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/CMP%20JS%20API%20v1.1%20Final.md

package.json
```diff
@@ -9,16 +9,20 @@
   },
   "devDependencies": {
     "autoprefixer": "^8.4.1",
+    "axios": "^0.18.0",
```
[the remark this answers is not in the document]
leshniak: minified axios has 12kb. We could use a bare XHR instead, what do you think?

src/index.js
```diff
@@ -12,51 +12,44 @@ const DEFAULT_OPTIONS = {
     cookieRejectExpiration: null,
     country: null, // country code
     countriesRequiringPrompt: null, // array of lower case country codes
+    customPublisherPurposes: [],
```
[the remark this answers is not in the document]
leshniak: ofc, I'll do that

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,388 @@
...
+
+        let vendorIds = [];
+
+        if (Array.isArray(ids)) {
+            vendorIds = ids;
+        } else {
+            vendorIds = maxVendorId ? createIdArray(1, maxVendorId) : vendorList.vendors.map(({id}) => id);
+        }
+
+        return new Promise((resolve) => resolve({
```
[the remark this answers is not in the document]
leshniak: I think so :( It's more readable with promises. I know, I have not added any polyfill yet...

src/ConsentManagementProvider.js
leshniak: Ok, I agree with that ;). I'll change them to regular functions.

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,388 @@
...
+import axios from 'axios';
+
+const CMP_ID = 141;
+const CMP_VERSION = 1;
+const CMP_DEFAULT_LANGUAGE = 'en';
+const VENDOR_CONSENT_COOKIE_NAME = 'euconsent';
+const PUBLISHER_CONSENT_COOKIE_NAME = 'eupubconsent';
+const MAX_STANDARD_PURPOSE_ID = 24;
+const MAX_CUSTOM_PURPOSE_ID = 88;
+const getDefaultCookieAttributes = () => ({
```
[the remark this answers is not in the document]
leshniak: I case of plain object we could use `Object.assign({}, DEFAULTS, ...)`, but in case of nested objects this pattern allows us to keep the defaults immutable across instances and do not modify them by accident.

src/ContentManagementProvider-test.js
leshniak: don't worry UT are in "to do" list ;)

package.json
leshniak: I have the same concerns, but just love decorators ;). But seriously I want to clearly mark some methods as available from the `__cmp` global which is a part of the IAB framework, as well as from an instance. And I want to avoid the explicit list of method names somewhere, because then you have to jump through the code to find out what is what.

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,388 @@
...
+        array.forEach((id) => {
+            map[Number(id)] = !!predicate(id);
+        });
+    }
+
+    return map;
+}
+
+class ConsentManagementProvider {
+    static installStub(gdprAppliesGlobally = false) {
```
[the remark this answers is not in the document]
leshniak: The case for it is that you should install the stub as soon as possible, to collect the calls for later. But you may not want to create an instance which implements full functionality yet. When things are ready, you will replace the stubbed `__cmp` with full-featured one, which is a part of an instance. I think that the static function is better in this case, because `installStub` has no sense as independent being without functions provided by the class.

src/ConsentManagementProvider.js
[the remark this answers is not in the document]
leshniak: Done

src/TrackingOptIn-test.js
```diff
@@ -1,22 +1,21 @@
...
 describe('TrackingOptIn', () => {
     let trackingOptIn;
     let tracker;
-    let optInManager;
     let geoManager;
-    let contentManager;
-    let onAcceptTracking;
-    let onRejectTracking;
+	let contentManager;
+	let consentManagementProvider;
```
leshniak: indentation

src/TrackingOptIn.js
```diff
@@ -62,41 +68,44 @@ class TrackingOptIn {
...
         this.render();
     }
 
-    clear() {
-        this.optInManager.clear();
-    }
-
     render() {
+        if (!this.geoRequiresTrackingConsent()) {
+            this.options.onUserConsent(true, null, null);
```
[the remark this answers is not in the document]
leshniak: Those are different things, `userConsent(false, [], [], [])` builds an object with `defaultConsent`, `allowedVendors`, `allowedVendorPurposes` and `allowedPublisherPurposes` for `TrackingOptIn.setConsents()` method which fires onUserConsent action while `this.options.onUserConsent(true, null, null)` is an external callback. But maybe I should name it differently if you are coinfused.

src/index.js
```diff
@@ -12,51 +12,46 @@ const DEFAULT_OPTIONS = {
...
     language: null,
     queryParamName: null,
     preventScrollOn: 'body',
     track: true,
     zIndex: 1000,
-    onAcceptTracking() {
```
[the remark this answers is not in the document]
leshniak: I agree that it breaks compatibility, but keep in mind that there is no such division in the IAB CMP model.

Basically what we get after user's action is a list of consents for particular _vendors _and _purposes_. So let's say that the user enables OpenX, Rubicon, IX and Beachfront, and other vendors are disabled. Which callback should we fire? `onAccept`? What if user opts-in only Rubicon and Beachfront? Should it be `onReject` now? Maybe we should filter those consents and fire both callbacks?

I know that our current architecture provides such division. But from the IAB CMP perspective, it has less sense. Also what I do in the App component is a preparation for the possible rendering of checkboxes for each vendor and purpose instead of global accept/reject (which can exist in the UI,  but more as a shortcut for enabling/disabling all elements).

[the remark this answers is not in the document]
nmonterroso: Whether or not there's such a division in any external model shouldn't have an impact on our side, especially since we are accepting all or accepting none. 

Once we actually work in the specific vendor checkmarks _then_ we can talk about making a major version bump and requiring apps to change their integration points, but I don't believe that time is now since from the integration perspective there are no functional changes.

src/TrackingOptIn.js
```diff
@@ -62,41 +68,44 @@ class TrackingOptIn {
...
 
-        switch (this.hasUserConsented()) {
-            case true:
-                this.options.onAcceptTracking();
-                break;
-            case false:
-                this.options.onRejectTracking();
-                break;
-            default:
+        return this.consentManagementProvider.fetchVendorList()
```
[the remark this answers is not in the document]
nmonterroso: Oops, realized this isn't the preact component. Bad naming on my part.

leshniak:
New version available.

src/index.js
```diff
@@ -4,6 +4,8 @@ import Tracker from "./Tracker";
 import ContentManager from "./ContentManager";
 import GeoManager from "./GeoManager";
 import TrackingOptIn from './TrackingOptIn';
+import ConsentManagementProvider from "./ConsentManagementProvider";
+import {getCookieDomain} from './utils';
```
nmonterroso: looks like this import is unused

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,367 @@
...
+import Cookies from 'js-cookie';
+import {getCookieDomain, getJSON} from './utils';
+
+const CMP_ID = 141;
+const CMP_VERSION = 1;
+const CMP_DEFAULT_LANGUAGE = 'en';
+const VENDOR_CONSENT_COOKIE_NAME = 'euconsent';
+const PUBLISHER_CONSENT_COOKIE_NAME = 'eupubconsent';
+const MAX_STANDARD_PURPOSE_ID = 24;
+const MAX_CUSTOM_PURPOSE_ID = 88;
```
nmonterroso: looks like this const is unused

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,367 @@
...
+        const {
+            gdprApplies,
+            gdprAppliesGlobally,
+            hasGlobalScope,
+            vendorList,
+            cookieAttributes,
+            defaultConsent
+        } = this.options;
+
+        this.gdprApplies = !!gdprApplies || !!gdprAppliesGlobally;
```
nmonterroso: is there a reason to keep these properties on the object _and_ in `this.options`? Which is the source of truth?

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,367 @@
...
+        this.gdprAppliesGlobally = !!gdprAppliesGlobally;
+        this.hasGlobalScope = !!hasGlobalScope;
+        this.defaultConsent = !!defaultConsent;
+        this.vendorList = vendorList;
+        this.cookieAttributes = Object.assign(getDefaultCookieAttributes(), cookieAttributes);
+    }
+
+    fetchVendorList(version) {
+        return getJSON(`https://vendorlist.consensu.org/${version ? `v-${version}/` : ''}vendorlist.json`);
```
nmonterroso: What happens when this promise rejects? Things still seem to work (accept/reject callbacks), but should we be reacting to the failed request at all?

src/ConsentManagementProvider.js
leshniak: Object prop has priority over options. These props are used many times across the methods, I did that for easier access and to avoid casting to Boolean in each method. Also I want to keep the originals untouched.

src/index.js
leshniak: yes, removed. some garbage from autoimport.

src/ConsentManagementProvider.js
leshniak: I think just throw an error because it's critical for the CMP. We could implement some kind of recovery if we would decide to add a vendor list to the bundle as a fallback.

nmonterroso: As long as the error doesn't prevent the page/lib from continuing to execute that's fine.

src/ConsentManagementProvider.js
```diff
@@ -0,0 +1,360 @@
...
+                        }
+                    };
+                    event.source.postMessage(returnMsg, '*');
+                });
+            }
+        }, false);
+    }
+
+    constructor(options) {
+        this.options = Object.assign(getDefaultOptions(), options);
```
nmonterroso: It's definitely strange to have duplicate configs as part of `this.options` and in `this.`. We should use either one or the other, otherwise it's confusing and unclear when one should be used over the other. 

For example, the `this.` properties are authoritative for everything listed below this line but `this.options` is authoritative for `allowedVendorPurposes` and the other properties in `createConsents`.

leshniak: So what I should do if I want to recalculate some part of options and store resulting values for the rest of the methods? Modify the original `options` object? Recalc same values over and over in each method where I want to use them? I'm thinking mainly about `this.gdprApplies` and `this.vendorList` because those have the biggest overhead. Dou you propose any design pattern?

leshniak: Hope you'll be satisfied after the last commit ;)

leshniak:
@nmonterroso if everything is ok now, please release our changes or give us a green light and we can do it ourselves ;)

nmonterroso:
@leshniak good to go. When you're ready to release use the [jenkins task](http://jenkins.wikia-prod:8080/view/CAKE/view/tracking-opt-in/job/tracking-opt-in/job/build,test,deploy/) and once the deploy is successful please post in #gdpr_engineering what the latest changes are.

nandy-andy:
Thanks @nmonterroso! We'll release it this week and update first on mobile-wiki :-)

PR #50, opened 2018-07-11
Increase timeout for browserstack

JoshuaRogan:
Increase the length of an individual test on browser stack before timing out. 

http://webdriver.io/guide/getstarted/configuration.html#waitforTimeout


nmonterroso:
LGTM 🤞

PR #51, opened 2018-09-17
PLATFORM-3678: GDPR cookies should sync after user clicks an accept/reject

themech:
GDPR cookies will be set on both domains (wikia and fandom) via the cookie-syncer service. This means that whenever users changes his mind about GDPR and sets new cookie values, we should sync this decision to the other domain in order to keep the values consistent.

FYI: @Wikia/cake


src/CookieSyncManager.js
```diff
@@ -0,0 +1,41 @@
...
+            '.fandom.com',
+            '.wikia-dev.pl',
+            '.fandom-dev.pl',
+            '.wikia-dev.us',
+            '.fandom-dev.us',
+        ];
+    }
+
+    getServicesDomain() {
+        return this.knownDomains.find(domain => this.host.endsWith(domain), this);
```
rogatty: `endsWith` isn't being transpiled, see https://github.com/babel/babel/issues/1739. I don't think we include the polyfill, so let's use something else.

src/TrackingOptIn-test.js
```diff
@@ -39,7 +42,8 @@ describe('TrackingOptIn', () => {
             onRejectTracking,
         };
 
-        trackingOptIn = new TrackingOptIn(tracker, optInManager, geoManager, contentManager, consentManagementProvider, options, window.location);
+        trackingOptIn = new TrackingOptIn(tracker, optInManager, geoManager, contentManager, consentManagementProvider,
```
rogatty: Break the line after `(` and before `)` instead. The same in lines below. Unfortunately, there is no linter :disappointed:

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@wikia/tracking-opt-in",
-  "version": "1.0.36",
+  "version": "1.0.37",
```
rogatty: This will be done as a part of http://jenkins:8080/job/tracking-opt-in/job/build,test,deploy/
Doing it manually can break something, I guess.

src/CookieSyncManager.js
```diff
@@ -0,0 +1,45 @@
...
+            '.wikia-dev.us',
+            '.fandom-dev.us',
+        ];
+    }
+
+    hostEndsWith(suffix) {
+        return suffix === this.host.slice(-suffix.length);
+    }
+
+    getTopLevelDomain() {
```
rogatty: Technically it's TLD+1 :stuck_out_tongue:

themech: `getServicesTopDomain` then? 😛

src/components/styles.scss
```diff
@@ -171,3 +171,7 @@ $desktop-devices: "only screen and (min-width: #{$breakpoint})";
     margin-bottom: 0; // save some space to prevent scroll bars
   }
 }
+
```
JoshuaRogan: 👍

src/TrackingOptIn.js
```diff
@@ -36,8 +38,18 @@ class TrackingOptIn {
             allowedVendors: this.options.enabledVendors,
             allowedVendorPurposes: this.options.enabledVendorPurposes
         });
-        this.consentManagementProvider.install();
         this.options.onAcceptTracking();
+
+        return this.consentManagementProvider.install();
```
nmonterroso: Is there a need to return this promise? Can you update the docs indicating the these functions return this promise?

rogatty: This is an internal, undocumented method. You probably think about the callback provided by users but it's being called above and didn't change.

src/components/styles.scss
rogatty: It's my favourite line too :wink:

themech:
changes look ok, let's proceed with it

PR #52, opened 2018-10-15
ADEN-7789 Updating vendor list

dmnsobczak:


src/OptInManager.js
```diff
@@ -9,6 +9,8 @@ export const STATUS = {
     ACCEPTED: 'accepted',
     REJECTED: 'rejected',
 };
+export const VERSION_COOKIE_NAME = 'tracking-opt-in-version';
+export const VERSION_CURRENT_ID = 2; // Increment to force modal again
```
nandy-andy: I was thinking of keeping here SemVer version we put in `package.json` but it doesn't give us anything, does it? If it doesn't give us anything it doesn't make sense to complicate things ;)

src/OptInManager.js
```diff
@@ -19,6 +21,14 @@ class OptInManager {
         this.queryParam = queryParam || DEFAULT_QUERY_PARAM_NAME;
     }
 
+    checkCookieVersion() {
+        const versionCookieValue = Cookies.get(VERSION_COOKIE_NAME);
+
+        if (!versionCookieValue || parseInt(versionCookieValue) < VERSION_CURRENT_ID) {
```
nandy-andy: I wanted to suggest passing a radix to `parseInt()` to make it more bullet-proof but maybe it's also over-engineering since [`VERSION_CURRENT_ID`](https://github.com/Wikia/tracking-opt-in/pull/52/files#diff-6d86632a62cd1951929821c001115732R13) is always an integer, right? ;-)

src/OptInManager.js
```diff
@@ -31,16 +41,19 @@ class OptInManager {
         return this.getValue() === STATUS.REJECTED;
     }
 
-    setTrackingAccepted() {
-        const attributes = {
-            expires: this.acceptExpiration,
-        };
-
+    setCookie(name, value, attributes = {}) {
```
nandy-andy: Shouldn't we rename it to `setCookies()` since it now sets two cookies :-)

src/OptInManager.js
dmnsobczak: Keeping SemVer here doesn't feel right, because we can fix some bugs, release new patch, bump version, but this might not need to reset people modal status. I would keep this as a accepted/rejected vendor list version and increase once we will change something in that logic.

src/OptInManager.js
dmnsobczak: It maybe is over-engineering, but why not? :D Changed.

src/OptInManager.js
dmnsobczak: You right, let's keep that correct :)

src/OptInManager-test.js
```diff
@@ -46,6 +51,19 @@ describe('OptInManager', () => {
         assert.equal(optInManager.getValue(), undefined);
     });
 
+    it('clears cookie on version change', () => {
+        let optInManager = new OptInManager();
+        optInManager.setTrackingAccepted();
+
+        Cookies.set(VERSION_COOKIE_NAME, VERSION_CURRENT_ID - 1);
```
jacekbj: It's counter-intuitive to test the decreasing value of current version, it would rather go up.

src/OptInManager-test.js
```diff
@@ -46,6 +51,19 @@ describe('OptInManager', () => {
         assert.equal(optInManager.getValue(), undefined);
     });
 
+    it('clears cookie on version change', () => {
+        let optInManager = new OptInManager();
+        optInManager.setTrackingAccepted();
```
jacekbj: What about a scenario when tracking is rejected?

src/OptInManager.js
```diff
@@ -19,6 +21,14 @@ class OptInManager {
         this.queryParam = queryParam || DEFAULT_QUERY_PARAM_NAME;
     }
 
+    checkCookieVersion() {
+        const versionCookieValue = Cookies.get(VERSION_COOKIE_NAME);
+
+        if (!versionCookieValue || parseInt(versionCookieValue, 10) < VERSION_CURRENT_ID) {
```
jacekbj: Please add a test scenario where cookie is not set.

src/OptInManager-test.js
dmnsobczak: I disagree. The idea is that value will increase in code (after new version release) and stay lower in cookie and that's what I'm mocking. I could overwrite constant in `OptInManager.class`, but I think that this approach is more production-consistent.

src/OptInManager-test.js
dmnsobczak: This scenario is exactly the same like for accepted - the only different thing is cookie expire time (one month instead of year) and opt-in cookie value (which is not tested here). Covering this scenario would be just copy-paste of existing code and reversing asserts values.

src/OptInManager.js
dmnsobczak: This test will become obsolete once we will pass the transition period, but more tests are never worse, right? :)

PR #55, opened 2018-10-29
ADEN-7869 GDPR check for mobile-app

pnather:


src/utils.js
```diff
@@ -37,3 +37,7 @@ export function getJSON(url) {
         req.send(null);
     });
 }
+
+export function urlContainsParam(url, param) {
+    return new URL(url).searchParams.get(param) === 'true';
```
dmnsobczak: Is it only string `true`? What about Boolean true and/or `1`? Please verify that with @gaperlinski.

src/index.js
```diff
@@ -114,6 +115,9 @@ export default function main(options) {
         },
         window.location,
     );
-    instance.render();
+
+    if (!urlContainsParam(window.location.href, 'mobile-app')) {
```
dmnsobczak: You expect to get some instance returned from function `main()`. Please check if this won't brake anything on app/mw/f2.

src/utils.js
```diff
@@ -37,3 +37,7 @@ export function getJSON(url) {
         req.send(null);
     });
 }
+
+export function urlContainsParam(url, param) {
```
dmnsobczak: I would rename it to `isParameterSet` and move here `window.location.href` to call it just `isParameterSet('mobile-app')`.

src/utils.js
pnather: they set always to 'true'

PR #57, opened 2018-10-30
ADEN-7869-3 GPDR on mobile

pnather:


src/TrackingOptIn-test.js
```diff
@@ -57,6 +57,39 @@ describe('TrackingOptIn', () => {
...
+
+        Object.defineProperty(window, 'location', {
+            value: windowLocation
+        });
+
+        trackingOptIn.render();
+
+        assert.isNotOk(modalIsShown());
+
+        delete window.location;
```
jacekbj: I know I'm a pain in the **** but this test checks something else than it's supposed to be from this point on :D

src/utils-test.js
```diff
@@ -4,6 +4,12 @@ import * as utils from './utils';
 const noop = () => {};
 
 describe('Utils', () => {
+    context('isParameterSet', () => {
+        it('checks if query parameter is set', () => {
+            assert.isNotTrue(utils.isParameterSet('parameter'));
```
jacekbj: What is the `window.location.href` set to?

PR #61, opened 2019-07-09
Cake-4597-2 | Updates to GDPR modal

cschucode:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "9.9.0"
+  - "10.7.0"
```
mklucsarits: Just wondering, how did you decide upon this version number? Is it matching the node version somewhere else?

cschucode: When doing the jenkins build, the error read "did you mean 10.7.0". I guess I did.

JoshuaRogan: If we want to upgrade the node version we have to do it in a few places. Unless there is a need to I would rather not.

PR #62, opened 2019-07-09
Run jenkins tasks on qa executors instead

frankfarmer:
The major dependency of this build is nodejs.  Changing nodejs version seems to trigger a g++ build.  g++ is installed on executors, but not docker-builder

so this build might be better suited to qa-executor nodes.


cschucode:
LGTM.

PR #68, opened 2019-07-26
Handle user input for vendor and purpose IDs

JoshuaRogan:


src/ConsentManagementProvider.js
```diff
@@ -336,11 +336,11 @@ class ConsentManagementProvider {
             consents = {
                 metadata: this.vendorConsent.getMetadataString(),
                 purposeConsents: toAllowedMap(
-                    createIdArray(1, MAX_STANDARD_PURPOSE_ID),
+                    createIdArray(1, MAX_STANDARD_PURPOSE_ID), // get from user
```
dmnsobczak: 👍

src/ConsentManagementProvider.js
```diff
@@ -336,11 +336,11 @@ class ConsentManagementProvider {
...
                     (id) => this.vendorConsent.isPurposeAllowed(id)
                 ),
                 vendorConsents: toAllowedMap(
-                    vendorIds,
+                    vendorIds, // get from user
```
dmnsobczak: 👍

src/OptInManager.js
```diff
@@ -52,6 +61,10 @@ class OptInManager {
 
         Cookies.set(name, value, attributes);
         Cookies.set(VERSION_COOKIE_NAME, CMP_VERSION, attributes);
+
+        // set a new cookie with the vendor list and purpose list as JSON
+        const cookieValue = {vendorIds: this.vendorIds, purposeIds: this.purposeIds};
+        Cookies.set(VENDOR_PURPOSE_LIST_COOKIE_NAME, JSON.stringify(cookieValue), attributes);
```
dmnsobczak: Do we really need a new cookie for that? Those values are stored inside the consent string:
https://github.com/Wikia/tracking-opt-in/blob/master/src/ConsentManagementProvider.js#L41

src/ConsentManagementProvider.js
JoshuaRogan: Hmm it looks like this PR in the current state properly saves the consent cookie. What is this part of the code for?

dmnsobczak: Ook, I got it now. Try opening the console on Communities and type:
`window.__cmp('getVendorConsents', null, (consentData) => console.log(consentData));`
You will call cmp wrapper directly and execute `getVendorConsents` function from L316.

PR #70, opened 2019-07-26
Build toggle component

vforge:
A minimal preact ios switch (styling borrowed from here and there)

### DIFF
--- a/src/components/Switch.js
+++ b/src/components/Switch.js
@@ -1,0 +1,26 @@
+import { h, Component } from 'preact';
+
+import styles from './styles.scss';
+
+class Switch extends Component {
+    onChange = (e) => {
+        e.preventDefault();
+
+        this.props.onChange(e.target.checked);
+    }
+
+    render({ isOn }) {
+        const id =  `switch_${Math.floor(Math.random() * 10000)}`;
+
+        return (
+            <div className={styles.switch} onChange={this.onChange}>
+                <input id={id} checked={isOn} type="checkbox" />
+                <label htmlFor={id} >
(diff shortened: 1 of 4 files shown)


src/components/Switch.js
```diff
@@ -0,0 +1,26 @@
...
+        e.preventDefault();
+
+        this.props.onChange(e.target.checked);
+    }
+
+    render({ isOn }) {
+        const id =  `switch_${Math.floor(Math.random() * 10000)}`;
+
+        return (
+            <div className={styles.switch} onChange={this.onChange}>
```
mklucsarits: What exactly is happening here with the CSS? You're importing styles from a stylesheet as a JS object, which is confusing in itself, but then using it as the component's `className` instead of as `style`. There must be some magic I'm not understanding here.

src/components/styles.scss
```diff
@@ -171,3 +173,58 @@ $desktop-devices: "only screen and (min-width: #{$breakpoint})";
...
+    }
+
+    label {
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        cursor: pointer;
+        width: $width;
+        height: $height;
+        background: grey;
```
mklucsarits: 🇬🇧

src/components/Switch.js
vforge: It's webpack's/react magic - see here https://programmingwithmosh.com/react/css-modules-react/

src/components/styles.scss
vforge: 🆗

PR #76, opened 2019-07-31
Separate styles for Switch

vforge:


src/components/Switch.scss
```diff
@@ -0,0 +1,67 @@
...
+        transform: translateX(-100%);
+    }
+
+    .label:active span {
+        background: $color-primary;
+        width: $height;
+    }
+}
+
+.preferencesSection {
```
mklucsarits: This shouldn't be in Switch.scss. I'll move this myself to Preferences.scss.

vforge: Sorry about it, I was too clumsy on copy/paste

PR #78, opened 2019-07-31
CAKE-5009 | Use sane defaults for vendors/purposes

JoshuaRogan:
Minor PR to use the defaults for the vendors/purposes. 

@Wikia/cake


cschucode:
Looks good. May need to merge with `gdpr-iab` again after my ScreenOne changes.

PR #80, opened 2019-08-01
Add purposes to preferences screen 

mklucsarits:
There are still several TODOs for the UI here.

### DIFF
--- a/src/components/Preferences.js
+++ b/src/components/Preferences.js
@@ -1,12 +1,29 @@
 import { h, Component } from 'preact';
 import PreferencesSection from './PreferencesSection';
 
-import styles from './styles.scss';
+import globalStyles from './styles.scss';
+import styles from './Preferences.scss';
 
 class Preferences extends Component {
     render(props, state) {
+        const { appOptions } = props;
         return (
-            Preferences
+            <div
+                className={globalStyles.overlay}
+                style={{
+                    zIndex: appOptions.zIndex,
+                }}
+            >
+                <div className={globalStyles.container}>
+                    <h2 className={`${styles.heading} ${styles.preferencesHeading}`}>Preference Settings</h2>
+                    <div className={styles.preferencesDescription}>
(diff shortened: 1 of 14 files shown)


src/components/Preferences.js
```diff
@@ -1,12 +1,94 @@
...
+    };
+
+    componentWillMount() {
+        if (!this.state.purposes) {
+            fetchVendorList().then((json) => {
+                // Filter purposes to those used by Fandom
+                const purposes = json.purposes.filter(purpose => (this.props.allPurposes.indexOf(purpose.id) >= 0));
+                // Filter vendors to those used by Fandom
+                const vendors = json.vendors.filter(vendor => (this.props.allVendors.indexOf(vendor.id) >= 0));
+                this.setState({ purposes });
```
JoshuaRogan: Should this be both purposes and vendors?

src/components/PreferencesSection.js
```diff
@@ -1,26 +1,39 @@
...
+    toggleIsExpanded() {
+        const { isExpanded } = this.state;
+        this.setState({ isExpanded: !isExpanded });
+        this.forceUpdate();
+    }
+
     render(props, state) {
 		const { heading, isEnabled } = props;
-		const { isExpanded } = state;
+        const { isExpanded } = this.state;
```
JoshuaRogan: If we are going to use `this.state` we should probably remove `state` from the method signature. Both work in preact but we should be consistent.

src/components/Preferences.js
mklucsarits: Yeah, I'm getting there. This file is still very WIP.

src/components/PreferencesSection.js
mklucsarits: Good point

src/components/Preferences.js
JoshuaRogan: Sounds good. Overall, I'm on board with this approach.

src/components/ScreenOne.scss
```diff
@@ -6,65 +6,82 @@
     }
 }
 
+.dialog {
+    background-color: white;
+    border-top-left-radius: 4px;
+    border-top-right-radius: 4px;
+    display: flex;
+    flex-direction: column;
+    font-family: Rubik, "Helvetica Neue", Helvetica, Arial, sans-serif !important;
```
vforge: There's a variable for it now

mklucsarits: It was actually duplicate code that I now removed.

PR #81, opened 2019-08-01
CAKE-5019 | Add all tracking events

vforge:


src/Tracker.js
```diff
@@ -12,6 +12,10 @@ const TRACK_PARAMS = {
 };
 const TRACK_TIMEOUT = 3000;
 
+const TRACKING_CATEGORY = 'gdpr-modal';
+const ACTION_IMPRESSION = 'Impression';
```
mklucsarits: Are you sure these should be capitalized? They're lowercase in the spreadsheet.

src/Tracker.js
```diff
@@ -77,6 +83,96 @@ class Tracker {
...
+    /**
+     * All tracking events
+     * @see https://docs.google.com/spreadsheets/d/1SdKVYFpAcW4xnLDOhLmpZ1h-Kr8Dq2a0tk-fb_GEEms/edit#gid=0
+     */
+    trackViewImpression() {
+        this.trackImpression('modal-view');
+    }
+
+    trackNoCookieImpression() {
+        this.trackImpression('no-coolor');
```
mklucsarits: Typo (this event happens when you forget to pack a cooler for your camping trip).

src/Tracker.js
vforge: That's the original piece of the code https://github.com/Wikia/tracking-opt-in/pull/81/files#diff-57aa11691094c3295e1f9dcbe8432ba2L8

But you're right, I'll downcase them and check with Keiko.

src/Tracker.js
vforge: omg :shy:

src/Tracker.js
mklucsarits: ok I see that's how they were originally so 🤷‍♂

vforge: They're lowercased in other products... 🤷‍♂

PR #82, opened 2019-08-02
Update i18n files, make sure we can replace text when necessary, fix styles

vforge:


src/ContentManager.js
```diff
@@ -26,10 +54,25 @@ export const langToContent = {
...
+                ...content,
+                ...langToContent[lang],
+            };
+        }
+
+        this.content = content;
+    }
+
+    t(key) {
+
```
JoshuaRogan: Is this the translate function that returns a string?

vforge: Oh, this should be removed, one sec

JoshuaRogan:
LGTM Tested on chrome

vforge:
@JoshuaRogan Again, I'll fix tests

src/components/App.js
```diff
@@ -88,7 +88,7 @@ class App extends Component {
                 <ScreenOne
                     appOptions={options}
                     content={content}
-                    text={content.bodyParagraphs}
+                    text={content.mainParagraphs}
```
mklucsarits: If `content` is already passed in to this component, I think we should just use it directly and remove this `text` prop.

vforge: thanks for catching it, I already removed this prop in the component

src/ContentManager.js
```diff
@@ -13,7 +13,35 @@ import ru from './i18n/ru/tracking-opt-in.json';
...
 
-export const langToContent = {
+/**
+ * Additional strings (URLs, mostly)
+ * - Keys for this object are the same as keys for `langToContent`
+ * - default values are specified in `en` key
+ */
+const additionalStrings = {
+    en: {
+        privacyPolicyUrl: "https://www.fandom.com/privacy-policy",
```
mklucsarits: Not a blocker but we should use single-quote strings.

vforge: doh, right

PR #83, opened 2019-08-02
CAKE-5006 Further UI done for preferences screen

mklucsarits:


src/components/PreferencesVendorList.scss
```diff
@@ -0,0 +1,70 @@
+@import "./consts.scss";
+
+.vendorList {
+    border: 1px solid #5f7a7b;
```
vforge: Can we make this color a variable?

src/components/PreferencesVendorList.scss
```diff
@@ -0,0 +1,70 @@
...
+.header {
+    background-color: #5f7a7b;
+    color: #fff;
+    font-size: 14px;
+    font-weight: 500;
+    padding: 4px 12px;
+}
+
+.subheader {
+    background-color: #7f9998;
```
vforge: And this one?

src/components/Preferences.js
```diff
@@ -38,35 +39,56 @@ class Preferences extends Component {
...
     }
 
-    togglePurpose(isEnabled) {
-        const { consentedPurposes, consentedVendors } = this.props;
-        // TODO this.props.updatePurposes()
+    togglePurpose(purposeId, isEnabled) {
+        const { consentedPurposes, consentedVendors, updatePurposes } = this.props;
+        if (isEnabled) {
+            if (consentedVendors.indexOf(purposeId) < 0) {
+                consentedVendors.push(puposeId);
```
JoshuaRogan: Is this allowed? It looks like you are mutating props.

mklucsarits: You're right, I didn't see a warning or error but I should definitely make copies of these arrays.

JoshuaRogan: Maybe it would make sense to be able to have a function `addConsentedVendor` so you can just pass the new one that needs to be added

JoshuaRogan: or `updateVendor(vendorId, enable)`

src/components/Preferences.js
```diff
@@ -38,35 +39,60 @@ class Preferences extends Component {
...
+        const { consentedPurposes, consentedVendors, updatePurposes } = this.props;
+        if (isEnabled) {
+            if (consentedVendors.indexOf(vendorId) < 0) {
+                const newConsentedVendors = consentedVendors;
+                newConsentedVendors.push(vendorId);
+                updatePurposes(newConsentedVendors, consentedPurposes);
+            }
+        } else {
+            const newConsentedVendors = consentedVendors;
+            newConsentedVendors.filter(id => (vendorId !== id));
```
cschucode: Does this line actually change the array, or would it need to be assigned?

mklucsarits:
I think I addressed all the comments so far, and I've added the chevron icon, so the UI portion is complete with this PR.

PR #85, opened 2019-08-05
CAKE-5019 Hook up remaining tracking events

mklucsarits:
Tracking events for new two-screen dialog

### DIFF
--- a/src/components/Preferences.js
+++ b/src/components/Preferences.js
@@ -11,10 +11,10 @@ function getParagraphs(blockOfText, content) {
             return content[key];
         }
         if (key === 'privacyPolicy') {
-            return `<a href=${content.privacyPolicyUrl} class=${globalStyles.link}>${content.privacyPolicyButton}</a>`;
+            return `<a href="${content.privacyPolicyUrl}" class="${globalStyles.link}" data-privacy-policy="true">${content.privacyPolicyButton}</a>`;
         }
         if (key === 'partnerList') {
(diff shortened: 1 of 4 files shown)


src/components/Preferences.js
```diff
@@ -110,7 +139,7 @@ class Preferences extends Component {
                 <div className={`${globalStyles.dialog} ${styles.dialog}`}>
                     <div className={styles.content}>
                         <h2 className={`${styles.heading} ${styles.preferencesHeading}`}>{content.preferencesHeadline}</h2>
-                        <div className={styles.preferencesDescription}>
+                        <div className={styles.preferencesDescription} onClick={(e) => this.clickDescription(e)}>
```
JoshuaRogan: You could use an instance field here instead if you prefer.

src/components/PreferencesSection.js
```diff
@@ -15,9 +15,27 @@ class PreferencesSection extends Component {
     }
 
     toggleIsExpanded() {
+        const { purpose, tracker } = this.props;
         const { isExpanded } = this.state;
         this.setState({ isExpanded: !isExpanded });
         this.forceUpdate();
+
+        switch (purpose.id) {
```
JoshuaRogan: Maybe turn this into a function `trackPurposeId(id, tracker)` as it's used twice 🤷‍♂

mklucsarits: The events/methods called for the two are different though.

JoshuaRogan: Ah okay. I didn't realize that

PR #88, opened 2019-08-05
Updating karma and webdriver to latest version

JoshuaRogan:


selenium/test.js
```diff
@@ -22,107 +22,110 @@ const cookieState = {
...
-    browser.setCookie({ name: 'Geo', value: '{' });
+    browser.setCookies({ name: 'Geo', value: '{' });
 }
 
 function ensureUserPrompt() {
-    browser.waitForExist(overlay);
-    assert(browser.isExisting(overlay));
-    assert(browser.isExisting(acceptButton));
-    assert(browser.isExisting(learnMoreButton));
+    browser.$('html').waitForExist(overlay);
```
cschucode: is browser required on these?

https://webdriver.io/docs/api/element/isExisting.html

JoshuaRogan: My understanding is that browser should be prefixed for all wdio commands.

PR #93, opened 2019-08-07
CAKE-5003 Design updates and fixes

mklucsarits:
* Simplified some markup + CSS
 * Implemented changes requested by Design
 * Reverted CMP cookie version

### DIFF
--- a/src/components/Preferences.js
+++ b/src/components/Preferences.js
@@ -11,10 +11,10 @@ function getParagraphs(blockOfText, content) {
             return content[key];
         }
         if (key === 'privacyPolicy') {
-            return `<a href="${content.privacyPolicyUrl}" class="${globalStyles.link}" data-privacy-policy="true">${content.privacyPolicyButton}</a>`;
+            return `<a href="${content.privacyPolicyUrl}" class="${globalStyles.link}" target="_blank" data-privacy-policy="true">${content.privacyPolicyButton}</a>`;
         }
         if (key === 'partnerList') {
-            return `<a href="${content.partnerListUrl}" class="${globalStyles.link}" data-partner-list="true">${content.partnerListButton}</a>`;
+            return `<a href="${content.partnerListUrl}" class="${globalStyles.link}" target="_blank" data-partner-list="true">${content.partnerListButton}</a>`;
         }
         return match;
     });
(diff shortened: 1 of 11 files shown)


src/i18n/tracking-opt-in.json
```diff
@@ -3,11 +3,11 @@
...
-        "For additional options, please select %learnMoreButton%. Note that if you do not accept, you may see ads that are less relevant to you and certain features of the site may not work as intended. You can change your mind and revisit your consent choices at any time by visiting our \"%privacyPolicyButton%\"."
+        "For additional options, please select \"%learnMoreButton%.\" Note that if you do not accept, you may see ads that are less relevant to you and certain features of the site may not work as intended. You can change your mind and revisit your consent choices at any time by visiting our %privacyPolicy%. For information about our partners using cookies on our site, please see the %partnerList%."
 ],
-    "acceptButton": "Accept",
+    "acceptButton": "ACCEPT",
     "backButton": "Back",
-    "learnMoreButton": "Learn More",
+    "learnMoreButton": "LEARN MORE",
```
JoshuaRogan: If we want this in all caps for all languages we might want to do this with `text-transform: uppercase` instead

mklucsarits: We're already doing that in the CSS for the `<button>`s it's used on, however this string is also used in some paragraph text and we'd like it uppercase there as well, at least for English. I suppose I could dynamically wrap it with a CSS class in that case also.

src/components/Preferences.js
```diff
@@ -11,10 +11,10 @@ function getParagraphs(blockOfText, content) {
             return content[key];
         }
```
cschucode: Does it make sense to have the same `getParagraphs()` **ScreenOne.js** and **Preferences.js**?

src/components/ScreenOne.js
```diff
@@ -8,13 +8,30 @@ function getParagraphs(blockOfText, content) {
         if (content[key]) {
             return content[key];
         }
```
cschucode: Same logic as **Preferences.js** `getParagraphs()`.

src/components/Preferences.js
mklucsarits: It doesn't make sense I know, but the function uses JSX so we can't move it to a non-JSX utility file (at least I don't think so).

src/i18n/tracking-opt-in.json
cschucode: It might be easier if we stayed consistent in this file with capitalization of buttons.

mklucsarits: I agree that probably would have been better. We've already resubmitted the translations once and would add more wait time to undo this change.

src/components/Preferences.js
cschucode: Actually, looks like it would be more trouble importing and exporting global styles etc.

PR #94, opened 2019-08-07
move getParagraphs into separate file

cschucode:


src/utils/getParagraphs.js
```diff
@@ -0,0 +1,21 @@
...
+        if (key === 'privacyPolicy') {
+            return `<a href="${content.privacyPolicyUrl}" class="${globalStyles.link}" target="_blank" data-privacy-policy="true">${content.privacyPolicyButton}</a>`;
+        }
+        if (key === 'partnerList') {
+            return `<a href="${content.partnerListUrl}" class="${globalStyles.link}" target="_blank" data-partner-list="true">${content.partnerListButton}</a>`;
+        }
+        return match;
+    });
+
+    return blockOfText.map(line => <p dangerouslySetInnerHTML={{ __html: replaceKeysInText(line) }} />);
```
JoshuaRogan: Should have commented this earlier but can we add a note that we know what we are doing here. whenever we use `dangerouslySet...` we should probably indicate why we need it.

cschucode: Left comments.

PR #99, opened 2019-08-08
CAKE-5003 Update some documentation

mklucsarits:


README.md
```diff
@@ -41,24 +41,24 @@ Invocation of the exported function returns an instance of `TrackingOptIn`. See
...
 - `language` - Override the language used to display the dialog text. Defaults to `window.navigator.language` if available, otherwise to `en`.
 - `preventScrollOn` - Prevent scrolling on the specified element when the dialog is shown. Can be either an element or query selector passed to `document.querySelector`. Defaults to `'body'`, set to `null` to prevent this behavior.
 - `queryParam` - The name of the query param to forcefully set the accepted status. Defaults to `tracking-opt-in-status` and accepts the string values `true` or `false` (e.g `http://starwars.wikia.com/?tracking-opt-in-status=true`)
-- `track` - whether to track impressions and user consent/rejections. Defaults to `true`.
+- `track` - Whether to track impressions and user consent/rejections. Defaults to `true`.
 - `zIndex` - Useful if elements on the app are appearing above the overlay/modal. Defaults to `1000`.
 - `onAcceptTracking` - The callback fired when:
```
JoshuaRogan: This needs to be updated. It will now always be called as you are always "accepting" from the perspective of this modal. You are only setting the `euconsent` cookie to disable particular vendors (maybe all)

mklucsarits: I added clarification on the 2nd bullet point but as it is, the conditions are still true.

mklucsarits: I also added to the Notes down below about our own (non-third-party) tracking.

PR #101, opened 2019-08-09
CAKE-5003 Prevent button styles from being overridden

mklucsarits:
Button styles were being overridden by wiki page styles.

### DIFF
--- a/src/components/Preferences.js
+++ b/src/components/Preferences.js
@@ -131,23 +131,22 @@ class Preferences extends Component {
                         {this.renderPreferenceSections(purposes)}
                     </div>
                     <div className={globalStyles.footer}>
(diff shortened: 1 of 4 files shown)


src/components/Preferences.js
```diff
@@ -131,23 +131,22 @@ class Preferences extends Component {
                         {this.renderPreferenceSections(purposes)}
                     </div>
                     <div className={globalStyles.footer}>
-                        <div className={globalStyles.buttons}>
```
vforge: Do we want to remove this wrapper?

mklucsarits: Yes, it's not needed anymore. I already removed it from the first screen but forgot it on the preferences screen.

PR #106, opened 2019-08-13
CAKE-5064 Allow rejection of non-IAB vendors

mklucsarits:
This covers the UI for rejecting tracking for non-IAB vendors.

### DIFF
--- a/src/components/OtherPartners.js
+++ b/src/components/OtherPartners.js
@@ -33,7 +33,7 @@ class OtherPartners extends Component {
 
         const toRender = NON_IAB_VENDORS.map((vendor) => {
             return (
-            <div className={vendorListStyles.vendor}>
+            <div className={vendorListStyles.vendor} key={vendor.name}>
                 <div className={vendorListStyles.flex}>
                     <div className={vendorListStyles.vendorName}>{vendor.name}</div>
                     <div>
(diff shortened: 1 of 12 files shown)


src/components/OtherPartners.js
```diff
@@ -0,0 +1,86 @@
...
+        onToggle(isOn);
+        tracker.trackOtherPartnersToggleClick();
+    }
+
+    renderVendors() {
+        const { content } = this.props;
+
+        const toRender = NON_IAB_VENDORS.map((vendor) => {
+            return (
+            <div className={vendorListStyles.vendor}>
```
JoshuaRogan: Do you need to add `keys` to these components?

mklucsarits: Probably, although there's no apparent number to use.

JoshuaRogan: It can be the name of the vendor. It doesn't need to be a number.

mklucsarits: I just added keys for every loop :)

JoshuaRogan: 🍠 I wonder if this has anything to do with requiring the `forceUpdate`. Maybe we aren't doing that elsewhere?

PR #116, opened 2019-08-15
CAKE-5003 Override link color from Fandom Creator communities

mklucsarits:
This was already done for other links but was missed in this file.


JoshuaRogan:
Good catch 💯

PR #118, opened 2019-08-22
ADEN-9245 Revert to using two callbacks for accept or reject

mklucsarits:
In order to accommodate existing logic on platforms using the GDPR dialog, we need to revert to using two separate callbacks from the dialog -- onAcceptTracking and onRejectTracking. These will be called when non-IAB vendor tracking is either accepted or rejected. The user-allowed IAB vendors and purposes will be passed as parameters to either callback as arrays.

Additionally I've fixed a race condition when initializing the CMP by utilizing an existing Promise.

### DIFF
--- a/src/TrackingOptIn-test.js
+++ b/src/TrackingOptIn-test.js
@@ -1,5 +1,5 @@
 import { assert } from 'chai';
-import { stub, createStubInstance, spy } from 'sinon';
+import { stub, createStubInstance } from 'sinon';
 import TrackingOptIn from './TrackingOptIn';
 import OptInManager from './OptInManager';
 import GeoManager from './GeoManager';
--- a/src/TrackingOptIn-test.js
+++ b/src/TrackingOptIn-test.js
@@ -1,5 +1,5 @@
 import { assert } from 'chai';
-import { stub, createStubInstance } from 'sinon';
+import { stub, createStubInstance, spy } from 'sinon';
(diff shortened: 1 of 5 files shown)


src/TrackingOptIn.js
```diff
@@ -114,6 +118,7 @@ class TrackingOptIn {
             preventScrollOn: this.options.preventScrollOn,
         };
 
+console.log('HERE 1', this.hasUserConsented());
```
nandy-andy: debugging?

src/TrackingOptIn-test.js
```diff
@@ -115,7 +118,9 @@ describe('TrackingOptIn', () => {
         trackingOptIn.render();
 
         assert.isNotOk(modalIsShown());
-        assert.isOk(onRejectTracking.called);
+        assert.isOk(consentManagementProvider.install.called);
+        // TODO assert function called after promise
+        // assert.isOk(onRejectTracking.called);
```
JoshuaRogan: would a timeout work?

JoshuaRogan: outside of that, I don't see any way besides properly setting up promise testing https://www.chaijs.com/plugins/chai-as-promised/

mklucsarits: I was overthinking. It's as easy as calling a stubbed function that resolves the promise and asserting the other function call with `then`.

PR #119, opened 2019-08-22
CAKE-5097 Don't use cutting Edge JS object

mklucsarits:
Replaces an object not supported by IE11. The `isParameterSet` function could instead use the same method, but I couldn't get window.location mocking to work that way for the mobile app unit test.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -34,6 +34,20 @@ export function isParameterSet(param) {
     return window.location.href.indexOf(`${param}=true`) !== -1;
 }
 
+export function getUrlParameter(paramName) {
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -34,6 +34,21 @@ export function isParameterSet(param) {
     return window.location.href.indexOf(`${param}=true`) !== -1;
 }
 
+export function getUrlParameter(paramName) {
+    const paramList = window.location.search.slice(1).split('&');
+    let paramValue = null;
+    paramList.forEach((param) => {
```
JoshuaRogan: `paramList.filter` might be a better approach but LGTM

PR #121, opened 2019-09-03
ADEN-9266 Disable CMP queue and implement optOut stub

dmnsobczak:


src/ConsentManagementProvider.js
```diff
@@ -244,24 +246,39 @@ class ConsentManagementProvider {
...
+                    queue.forEach((args) => cmp(...args));
+                });
+            } catch (error) {
+                void(error);
+                window.__cmp = cmp;
+                console.error(new Error('incompatible stub, cannot run queue'));
+            }
         }
+
+        this.options.mounted = true;
```
fraszczakszymon: It is really `option`? Should we keep that flag in some `state` object?

dmnsobczak: You right! Fixed.

PR #128, opened 2019-10-11
CAKE-5211 | Add isCurse param for privacy policy link

JoshuaRogan:
@Wikia/cake

### DIFF
--- a/src/TrackingOptIn-test.js
+++ b/src/TrackingOptIn-test.js
@@ -26,7 +26,7 @@ describe('TrackingOptIn', () => {
     }
 
     function isCurseUrl() {
-        return document.querySelector('a[href*=curse]');
+        return document.querySelector('a[href*=curse-privacy-policy]');
     }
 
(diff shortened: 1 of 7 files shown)


src/TrackingOptIn-test.js
```diff
@@ -25,6 +25,10 @@ describe('TrackingOptIn', () => {
         return document.querySelector(`.${styles.overlay}`);
     }
 
+    function isCurseUrl() {
+        return document.querySelector('a[href*=curse]');
```
mklucsarits: I think this selector should be more specific and include more or all of the URL, but that's just a suggestion.

JoshuaRogan: Good call. I switched it to `a[href*=curse-privacy-policy]` 

https://github.com/Wikia/tracking-opt-in/pull/128/commits/5a95a4daaa20ddb9d2b3495b8bc686042f0002f5

PR #129, opened 2019-11-14
ADEN-9580 Introducing CCPA basic mechanism

dmnsobczak:


src/ccpa/UserSignalMechanism.js
```diff
@@ -0,0 +1,213 @@
...
+});
+
+function createPrivacyString(explicitNotice = USP_VALUES.na, optOutSale = USP_VALUES.na) {
+    return `${USP_VERSION}${explicitNotice}${optOutSale}`;
+}
+
+class UserSignalMechanism {
+    mounted = false;
+
+    static installStub() {
```
Bielik20: I would move queueing functionality to separate file and reuse it across GDPR and CCPA.

src/ccpa/UserSignalMechanism.js
```diff
@@ -0,0 +1,213 @@
...
+            } else if (commandName === 'getQueue') {
+                callback(queue, true);
+            } else {
+                queue.push([commandName, version, callback]);
+            }
+        };
+    }
+
+    constructor(options) {
+        this.options = Object.assign(getDefaultOptions(), options);
```
Bielik20: I assume there is no spread operator support on the platform?

src/ccpa/UserSignalMechanism.js
```diff
@@ -0,0 +1,213 @@
...
+
+        console.log('CCPA: User Signal Mechanism initialized');
+    }
+
+    configure(options) {
+        Object.assign(this.options, options);
+    }
+
+    install() {
+        if (this.mounted) {
```
Bielik20: Is there a scenario when this would happen?

dmnsobczak: Right now: I think no.

src/ccpa/UserSignalMechanism.js
dmnsobczak: 👍

PR #135, opened 2019-12-11
ADEN-9619 Fixing undefined version issue

dmnsobczak:


src/ccpa/UserSignalMechanism.js
```diff
@@ -271,7 +271,7 @@ class UserSignalMechanism {
         return new Promise((resolve, reject) => {
             if (isCorrectVersion) {
                 resolve({
-                    version,
+                    version: version || USP_VERSION,
```
jacekbj: Why not `getUSPData(version = USP_VERSION)`?
BTW `version` can be a `string` as well as a `number`?

PR #149, opened 2020-05-12
ADEN-10225 Remove euconsent-v2 cookie if GVL tcfPolicyVersion changes.

jacekbj:
Move body of gatherTCF2Consent method into a Promise body,
which is resolved once vendor-list.json is fetched.

The vendor-list.json is cached for a week so we're safe
to postpone execution until it's downloaded.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -77,8 +77,6 @@ function initializeGDPR(options) {
 
     if (optInManager.checkCookieVersion()) {
         consentManagementProviderLegacy.setVendorConsentCookie(null);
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -77,8 +77,6 @@ function initializeGDPR(options) {
 
     if (optInManager.checkCookieVersion()) {
         consentManagementProviderLegacy.setVendorConsentCookie(null);
-        // ToDo: add additional reset if GVL will change
```
dmnsobczak, line 80: We can reimplement that in the future for TCF v2.0

PR #151, opened 2020-05-13
ADEN-10229 Adding translations mechanism

dmnsobczak:


src/modal/Preferences.js
```diff
@@ -61,6 +74,7 @@ class Preferences extends Component {
             updatePurposes(consentedVendors, newConsentedPurposes);
         }
 
+        // ToDo: fix purposes tracking
```
Bielik20, line 77: Shouldn't this be a ticket instead of a comment?

src/shared/ContentManager.js
```diff
@@ -57,15 +96,27 @@ export const langToContent = processLanguages({
 });
 
 export default class ContentManager {
-    lang = en;
     content = null;
+    language = 'en';
+
+    /**
+     * @returns Promise<VendorList|null>
```
Bielik20, line 103: Is it correct or remnant from copying?

src/shared/ContentManager.js
```diff
@@ -57,15 +96,27 @@ export const langToContent = processLanguages({
...
+     */
+    static fetchTranslation(language) {
+        if (!availableTranslations.includes(language)) {
+            return Promise.resolve(null);
```
Bielik20, line 107: Wouldn't it be better to always return actual translation?

Then you could get rid of that check: 
```js
if (translation) {
                    json.features = translation.features;
                    json.purposes = translation.purposes;
                    json.specialFeatures = translation.specialFeatures;
                    json.specialPurposes = translation.specialPurposes;
                }
```

src/shared/ContentManager.js
```diff
@@ -57,15 +96,27 @@ export const langToContent = processLanguages({
...
+    }
 
     constructor(lang) {
-        this.lang = lang;
+        this.language = lang;
 
         // all the strings default to `en`
         let content = langToContent.en;
 
+        // ToDo: translate "special" headers
```
Bielik20, line 119: Shouldn't this be a ticket instead of a comment?

src/modal/Preferences.js
dmnsobczak: Ticked added and linked:
https://wikia-inc.atlassian.net/browse/ADEN-10261

src/shared/ContentManager.js
dmnsobczak: Ticket added and linked:
https://wikia-inc.atlassian.net/browse/ADEN-10262

src/shared/ContentManager.js
dmnsobczak: No, because default, english translation is available only in full Vendor List. I don't think if moving Vendor List fetching to ContentManager is a good idea, at least for now. We have two different Vendor Lists for two TCF versions and we don't want to mix to much with an old version, even at cost of lower quality code for the transition stage. Moving only one version can introduce some confusion I guess. But it's definitely worth doing at last step - cleanup. Do you agree? Or should I move it now?

src/shared/ContentManager.js
dmnsobczak: It's fine. Translation file is a kind of Vendor List, but without vendors. Or maybe I misunderstood you?

PR #194, opened 2021-01-18
ADEN-10781 TCF v2.1 update

dmnsobczak:


src/modal/PreferencesVendorList.js
```diff
@@ -154,6 +155,21 @@ class PreferencesVendorList extends Component {
...
+                    {vendor.usesNonCookieAccess !== undefined && (
+                        <div className={styles.vendorDetail}>
+                            {vendor.usesNonCookieAccess ? content.storageDetailsNonCookieAccessTrue : content.storageDetailsNonCookieAccessFalse}
+                        </div>
+                    )}
+                    {vendor.cookieMaxAgeSeconds !== undefined && (
+                        <div className={styles.vendorDetail}>
+                            {content.storageDetailsCookieMaxAge} {getCookieAge(vendor.cookieMaxAgeSeconds, content)}
+                        </div>
+                    )}
```
Loczi94, line 171: I am not sure about this logic. 🤔  I think we should check the value of `usesNonCookieAccess` and if true display `storageDetailsNonCookieAccessTrue` and then check the value of `cookieMaxAgeSeconds` and if it is not undefined then display `storageDetailsNonCookieAccessFalse` and `storageDetailsCookieMaxAge` + time. Please, correct me if I am wrong.

dmnsobczak: Discussed on Zoom 👍

Loczi94 (approved):

PR #220, opened 2021-11-04
ADEN-11287 Introduce Fandom/Galactus into the modal

dmnsobczak:


src/modal/VendorDetails.js
```diff
@@ -0,0 +1,144 @@
+import { h, Component } from 'preact';
```
preludek, line 1: Is it necessary to import Component in preact if we are not using `extends`?

src/modal/VendorDetails.js
```diff
@@ -0,0 +1,144 @@
...
+import Consented from './utils/isConsented';
+
+class VendorDetails {
+    static renderVendor(vendor, props, hiddenSections = []) {
+        const { content, allPurposes, allPurposesSpecial, allFeatures, allFeaturesSpecial } = props;
+
+        return (
+            <div className={styles.vendorDetails}>
+                {vendor.purposes.length > 0 && (
+                    <div>{/* fragment */}
```
preludek, line 14: Won't an empty tag `<>` work the same or is it just Preact thing to use {/* fragment */}? 🤔  And if so, why not just import Fragment?

src/modal/VendorDetails.js
```diff
@@ -0,0 +1,144 @@
+import { h, Component } from 'preact';
+
+import styles from './PreferencesVendorList.scss';
+import getCookieAge from './utils/getCookieAge';
+import Consented from './utils/isConsented';
+
+class VendorDetails {
+    static renderVendor(vendor, props, hiddenSections = []) {
```
preludek, line 8: It seems for me that in this repo all classes returning some HTML (JSX, whatever :D) are extending Component. Not sure if this is a pattern in this repo, maybe we should (ask and) just stick to it?

src/modal/VendorDetails.js
```diff
@@ -0,0 +1,144 @@
...
+        const { content, allPurposes, allPurposesSpecial, allFeatures, allFeaturesSpecial } = props;
+
+        return (
+            <div className={styles.vendorDetails}>
+                {vendor.purposes.length > 0 && (
```
preludek, line 13: Just out of curiosity, is there a chance any of those `vendor.XXX` lines blow up? Due to a missing `purposes` key or it being undefined or smth?

src/modal/InternalUse.js
```diff
@@ -0,0 +1,101 @@
...
+
+    toggleIsVendorExpanded(vendorId) {
+        const { tracker, vendors } = this.props;
+        vendors.forEach((vendor) => {
+            if (vendor.id === vendorId) {
+                vendor.isExpanded = !vendor.isExpanded;
+            }
+        });
+        this.setState({ vendors });
+        this.forceUpdate();
```
preludek, line 34: is this also necessary? setState should trigger re-render.

preludek (commented):
Submited some questions :)

src/modal/VendorDetails.js
dmnsobczak: Actually not, thanks! Removed.

src/modal/VendorDetails.js
dmnsobczak: Actually this is a comment and it's not needed. Removed.

src/modal/VendorDetails.js
dmnsobczak: Those classes are using `state`, this one is just a simple static class used to encapsulate some part of the code and we can return renderable content by importing unused `h` from `preact`

src/modal/VendorDetails.js
dmnsobczak: Yes, this is a list of purposes for currently rendered vendor. Some vendors might have zero purposes but some special features and vice versa. In this case we won't render this section.

src/modal/InternalUse.js
```diff
@@ -0,0 +1,101 @@
...
+import sectionStyles from './PreferencesSection.scss';
+import vendorListStyles from './PreferencesVendorList.scss';
+
+class InternalUse extends Component {
+    state = {
+        isExpanded: false,
+    };
+
+    toggleIsInternalUseExpanded() {
+        const { tracker } = this.props;
```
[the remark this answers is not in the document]
dmnsobczak: Most of the code is copied or reused from other modules and this project was not originally written by AdEng. I fitted into the coding convention and I prefer to make bigger refactors in the separate ticket - I think we have a dev improvements ticket for that 🤔

src/modal/InternalUse.js
```diff
@@ -0,0 +1,101 @@
...
+        const { isExpanded } = this.state;
+        this.setState({ isExpanded: !isExpanded });
```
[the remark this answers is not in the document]
dmnsobczak: Most of the code is copied or reused from other modules and this project was not originally written by AdEng. I fitted into the coding convention and I prefer to make bigger refactors in the separate ticket - I think we have a dev improvements ticket for that 🤔

src/modal/InternalUse.js
dmnsobczak: Most of the code is copied or reused from other modules and this project was not originally written by AdEng. I fitted into the coding convention and I prefer to make bigger refactors in the separate ticket - I think we have a dev improvements ticket for that 🤔

PR #224, opened 2021-12-15
ADEN-11439 Even better GDPR for Xmas

nandy-andy:
We don't want to set any cookies without user permission. This implementation moves the responsibility for setting up session cookies to tracking-opt-in.


src/shared/CookieManager.js
```diff
@@ -0,0 +1,53 @@
...
+            ...cookie,
+            value: this.getSessionCookiesValue(cookie.name, cookie.addTimestamp)
+        }));
+    }
+
+    getSessionCookiesValue(name, addTimestamp) {
+        let resultValue = Cookies.get(name);
+
+        if (!resultValue && this.cache[name]) {
+            resultValue = this.cache[name];
```
preludek, line 18: I think this case will never happen. We are setting cache in the next if and since cookie names cannot be duplicated (as far as I checked, maybe I' wrong? 🙈 ) we will never get here.

src/shared/CookieManager.js
```diff
@@ -0,0 +1,53 @@
...
+        this.domain = getCookieDomain(window.location.hostname);
+        this.cache = {};
+        this.sessionCookies = cookies.map((cookie) => ({
+            ...cookie,
+            value: this.getSessionCookiesValue(cookie.name, cookie.addTimestamp)
+        }));
+    }
+
+    getSessionCookiesValue(name, addTimestamp) {
+        let resultValue = Cookies.get(name);
```
preludek, line 15: I don't think we have to use let here, this value doesn't need to be overwritten. You can const it and then just return it inside each if.

preludek (changes requested):

src/shared/CookieManager.js
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        return resultValue;
+    }
+
+    generateValue(name, withTimestamp) {
+        // for users from GDPR countries if they did not give consent for tracking
+        // we assign random values to session cookies; should match results of:
+        // https://developer.fastly.com/reference/vcl/functions/randomness/randomstr/
+        let resultValue = (Math.random().toString(36) + '_________________').slice(2, 12)
```
preludek, line 32: Also, do we need this weird dash part in the middle? `Math.random().toString(36).slice(2, 12)` does the same thing

nandy-andy: The `base64` generated by `Math.random().toString(36)` may be shorter because `0` will be removed, so we'll add `_` in those places and this way we'll match the characters from the function returned in VCL (`a-zA-Z0-9-_`) 😊

preludek (approved):

PR #225, opened 2021-12-20
ADEN-11445 Integrating Post-Quecast event mechanism

dmnsobczak:


src/index.js
```diff
@@ -112,8 +119,61 @@ function initializeCCPA(options) {
...
+
+    optInInstances.gdpr = initializeGDPR(options);
+    optInInstances.ccpa = initializeCCPA(options);
+
+    return optInInstances;
+}
+
+const autostartModal = () => {
+    if (!window.trackingOptInManualStart) {
+        // ToDo: move to AdEngine loader
```
nandy-andy, line 163: Let's remove this comment - nobody outside Fandom knows what AdEngine is and therefore the comment is just more confusing for people who get to the repo via npmjs.org 🙈

src/index.js
```diff
@@ -112,8 +119,61 @@ function initializeCCPA(options) {
...
+        window.trackingOptInOptions = {
```
nandy-andy, line 164: Why do we need it here? The application that uses `tracking-opt-in` passes the settings to the `main()` function when it starts. So, it should be up to the application to run the lib as soon as possible with the right options that are described in the README: https://www.npmjs.com/package/@wikia/tracking-opt-in#options

src/index.js
dmnsobczak: This is a note to the future self. This PR is marked as WIP, because it's merge will be blocked by a follow up ticket solving the problem of injecting TrackingOptIn library URL on the platform. It can be done by just adding it to the platform code or by finishing the AdEngine Loader PoC. After that those params will be removed from here.

src/index.js
dmnsobczak: Described above. Those options will be copied, hopefully to the UCP AdEngine Loader configuration. For now I left it for testing purposes.
ReadMe is already outdated, but after this PR running `main()` will be deprecated - TrackingOptI will be a self-starting application, just like AdEngine is.

src/index.js
nandy-andy: Does it mean we won't merge it after the QA but after the follow-up ticket is done (either platform code change or finishing AdEngine loader)?

src/index.js
nandy-andy: It's a perfect moment tu update the README file if you found it already outdated especially when you've added some changes to it within this PR ;) Are we sure we want to take away the ability to decide when to run `main()` from the platforms' creators?

src/index.js
dmnsobczak: Yes. We are in the code freeze so it either way will way for January and I might be able to push AdEngine Loader ticket further till this time :)

src/index.js
dmnsobczak: ReadMe is up-to-date with the first version of TrackingOptIn compliant with TCFv1. I would rather remove it and use Confluence instead.
Yes, modal should be loaded as fast as possible, especially after our changes with cookies. I'm not removing an ability to run it manually, it's just not recommended :D

nandy-andy: Modal should be loaded as fast as possible in the Fandom scenario. Maybe there are some scenarios and people who use the lib and they want to have full control. But maybe not - it has only <200 downloads this week.

Well, then maybe we should make the repository private and remove it from npmjs.org 🤔

nandy-andy (approved):
It's OK if we want just to test it locally and then change a bit again while introducing the new version to our platforms.

src/gdpr/ConsentManagementPlatform.js
```diff
@@ -55,9 +56,36 @@ class TrackingOptIn {
         });
         this.consentManagementProvider.install().then(() => {
             this.options.onRejectTracking(allowedVendors, allowedPurposes);
+            this.options.onConsentsReady();
         });
     };
 
+    getConsent() {
```
nandy-andy, line 63: Consider renaming to `getGDPRCOnsent()` as it clearly returns an object defining the GDPR consent.

README.md
```diff
@@ -75,8 +75,8 @@ The following options are accepted:
 - As of v2.0.0, accepting or rejecting _vendor tracking_ should not affect any GA or internal tracking unrelated to advertising.
 - Country codes are in [ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1) format.
 
-### TrackingOptIn class
-Calling the exported function returns an instance of the [`TrackingOptIn`](https://github.com/Wikia/tracking-opt-in/blob/master/src/TrackingOptIn.js) class. The class has the following functions:
+### ConsentManagementPlatform class
+Calling the exported function returns an instance of the [`ConsentManagementPlatform`](https://github.com/Wikia/tracking-opt-in/blob/master/src/TrackingOptIn.js) class. The class has the following functions:
 
 - `hasUserConsented()` - Returns `true` if the user has accepted _non-IAB_ vendor tracking (or does not need to based on their geo), `false` if they have explicitly rejected tracking, and `undefined` if the user has neither accepted nor rejected tracking.
```
nandy-andy, line 81: Since you're updating the docs it's worth expand section about options as well as update the section about the class you've just renamed.

nandy-andy (dismissed):
I left two small comments (easy to address, I hope 🤞). We also need to verify if the tests pass locally and if not, fix them and then create a follow-up to fix Travis or replace it with Jenkins maybe 🤔 (//cc @soniackozubek)

src/gdpr/ConsentManagementPlatform.js
dmnsobczak: Well, this method is used internally in TrackingOptIn and can be used externally by calling a correct instance. This object have structure like this: `instances: { gdpr: ConsentManagementPlatform::class, ccpa: UserSignalMechanism::class }` so it's hard to use wrong method. Also one is `getConsent()` and the other one is `getSignal()`, because that's compliant with IAB name convention. TL;DR: Adding additional GDPR there is a overcomplication. Additionally: I try to use as less of `GDPR` acronym as possible, because modal implements TCF rules, not GDPR act directly 🤔

README.md
dmnsobczak: To not block this ticket I created a follow up to update not just README, but all docs in general:
https://fandom.atlassian.net/browse/ADEN-11493
We can discuss with the team who should do that, because I know this project and somebody else updating this file might lead to the bus factor increase :)

dmnsobczak:
> I left two small comments (easy to address, I hope 🤞). We also need to verify if the tests pass locally and if not, fix them and then create a follow-up to fix Travis or replace it with Jenkins maybe 🤔 (//cc @soniackozubek)

For some reason Travis gets 403 when it requests wikia artifactory. I updated the lock file.

nandy-andy (approved):

PR #226, opened 2022-01-04
ADEN-11443 set tracking cookies on user consent

jakubbanas:


src/gdpr/TrackingOptIn.js
```diff
@@ -176,6 +180,41 @@ class TrackingOptIn {
...
+        const pvNumberGlobal = Cookies.get('pv_number_global');
+
+        return {
+            pvNumber: pvNumber ? parseInt(pvNumber, 10) : 0,
+            pvNumberGlobal: pvNumberGlobal ? parseInt(pvNumberGlobal, 10) : 0,
+            sessionId: sessionId || uuidv4(),
+        };
+    }
+
+    getDomain() {
```
mfarul, line 207: we should have a unit test here for domains 1 .. 5 parts

mfarul:
I see we unify tracking cookies here. What about tracking functions? I see we have at least 2 mechanisms i n our products a) direct push to DW or via window.dataLayer.push. We should encapsulate both here.

src/gdpr/TrackingOptIn.js
```diff
@@ -1,7 +1,9 @@
 import { h, render } from 'preact/dist/preact';
 import Modal from '../modal/Modal';
-import { isParameterSet, parseUrl } from '../shared/utils';
-import { API_STATUS } from './ConsentManagementProvider';
+import {debug, isParameterSet, parseUrl} from '../shared/utils';
```
nandy-andy, line 3: Do we need importing `debug` here? It seems it's not being used anywhere.

nandy-andy (approved):

PR #229, opened 2022-01-26
ADEN-11520 DW tracking methods

mfarul:


src/gdpr/ConsentManagementPlatform.js
```diff
@@ -1,14 +1,11 @@
-import { h, render } from 'preact/dist/preact';
```
dmnsobczak, line 1: I think importing this unused `h` is required, because without this package is not building/starting 🤔

src/gdpr/Tracker.js
```diff
@@ -1,43 +1,35 @@
-import Cookies from 'js-cookie';
 import { debug } from '../shared/utils';
+import TrackingEventsQueue from "../tracking/TrackingEventsQueue";
```
dmnsobczak, line 2: I miss Prettier here: some imports are quoted with `""` and some of them with `''` :(

src/index.js
```diff
@@ -7,12 +7,14 @@ import OptInManager from './gdpr/OptInManager';
 import Tracker from './gdpr/Tracker';
 import ConsentManagementPlatform from './gdpr/ConsentManagementPlatform';
 import UserSignalMechanism from './ccpa/UserSignalMechanism';
-import CookieManager from './shared/CookieManager';
+import CookiesBaker from './tracking/CookiesBaker';
```
dmnsobczak, line 10: ❤️

src/index.js
```diff
@@ -7,12 +7,14 @@ import OptInManager from './gdpr/OptInManager';
...
 import { communicationService } from './shared/communication';
 import { debug } from './shared/utils';
+import EventsTracker from "./tracking/EventsTracker";
```
dmnsobczak, line 13: ditto

src/index.js
```diff
@@ -106,7 +108,7 @@ function initializeCCPA(options) {
     const {
         test,
         ...depOptions
-    } = Object.assign({}, DEFAULT_CCPA_OPTIONS, options);
+    } = options;
```
dmnsobczak, line 111: I'm not sure if extracting options merging outside the method is a good idea, because it's easier to forget what elements should be merged together, but as long as it works as previously I'm fine with it 🤷

src/index.js
```diff
@@ -125,6 +127,10 @@ export default function main(options) {
 
     debug('MODAL', 'Library loaded and started');
 
+    const mergedOptions = Object.assign({}, DEFAULT_OPTIONS, options);
```
dmnsobczak, line 130: Can we call it `gdprMergedOptions`? Those are not common for both instances.

src/index.js
```diff
@@ -134,27 +140,34 @@ export default function main(options) {
...
-            ...optInInstances.gdpr.getConsent(),
+            ...gdprConsent,
             ...optInInstances.ccpa.getSignal(),
         });
         communicationService.dispatch({
             type: instancesAction,
             ...optInInstances,
         });
+        // this will also handle CCPA region as it is not GDPR one
+        const allowedToTrack = gdprConsent.geoRequiresConsent === false || gdprConsent.gdprConsent === true;
```
dmnsobczak, line 161: `gdprConsent.gdprConsent` is `true` when geo not requires consent - we should avoid checking geo if not necessary, because GDPR logic might change in time.
```
        if (!this.geoRequiresTrackingConsent()) {
            return {
                gdprConsent: true,
                geoRequiresConsent: false,
            };
        }
```

src/shared/utils-test.js
```diff
@@ -48,6 +48,18 @@ describe('Utils', () => {
 
             assert.equal(domain, '.somethingkiwi.co.nz');
         });
+
+        it('should return undefined for domain without extension', () => {
+            expect(utils.getCookieDomain('fandom')).to.equal(undefined);
```
dmnsobczak, line 53: Just an observation that we are using `assert.equal` everywhere 🤔

src/tracking/CookiesBaker.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    setOrExtendCookies(cookieValues) {
+        this.cookies.forEach(cookie => {
+            let value = cookieValues[cookie.name] || Cookies.get(cookie.name);
+
+            if (!value && cookie.value) {
+                value = typeof(cookie.value) === 'function' ? cookie.value() : cookie.value;
+            }
+            if (value) {
```
dmnsobczak, line 22: `if (value) else if (cookie.value)`?

src/tracking/DataWarehouseEventsSender.js
```diff
@@ -0,0 +1,33 @@
+const TRACKING_BASE_URL = 'https://beacon.wikia-services.com/__track/';
+const TIMEOUT_MS = 3000;
+
+function getEventPath(name) {
+    name = !name ? 'view' : name.toLowerCase();
```
dmnsobczak, line 5: If you setting and returning `view` if `name` is not set, then why not:
`if (!name) return 'view`;`? 🤔

src/tracking/EventsTracker-test.js
```diff
@@ -0,0 +1,94 @@
...
+
+        // when
+        tracker.startTracking(true, cookieJar)
+
+        // then
+        assert.isOk(sender.send.called);
+        assert.isOk(sender.send.calledWithMatch(
+            sinon.match.has('env', 'dev').and(
+                sinon.match.has('beacon', 'beacon')).and(
+                    sinon.match.has('pv_unique_id')))
```
dmnsobczak, line 52: This will be a nightmare if you are matching 5+ elements 🙈

src/gdpr/ConsentManagementPlatform.js
mfarul: hmm, strange yarn build was successful on my side and IDE was pointing that h was not used ...

src/index.js
mfarul: changed to new TRACKING_OPTIONS

src/index.js
mfarul: added new method, just want to make sure we undertand this behavior as it is hidden quite deep ;)

src/shared/utils-test.js
mfarul: added new helper and() method

src/tracking/EventsTracker-test.js
mfarul: added new and() method

src/tracking/CookiesBaker.js
mfarul: nope - but a good catch, new version made this clear i hope

src/shared/utils-test.js
mfarul: what can i say, just copied this from existing test when moved some code :(

dmnsobczak (approved):

README.md
```diff
@@ -1,6 +1,6 @@
 # tracking-opt-in
 
-FANDOM's gdpr opt-in dialog prompt.
+FANDOM's gdpr opt-in dialog prompt and FANDOM's event based DWH tracking library.
```
nandy-andy, line 3: What do you think of closing the repository and removing it from [npmjs.org](https://www.npmjs.com/package/@wikia/tracking-opt-in)? Over the years many different CMPs appeared on the market and this one is more and more connected to Fandom. I'm only not sure if removing it from the public registry is a simple way if we shouldn't inform whoever downloaded it with an advance 🤔

src/index.js
```diff
@@ -119,11 +126,18 @@ function initializeCCPA(options) {
     return userSignalMechanism;
 }
 
+function isAllowedToTrack(gdprConsent, ccpaOptions) {
+    // this will also cover non GDPR regions and CCPA as well, as they all will have this set to true
+    return gdprConsent.gdprConsent === true;
```
nandy-andy, line 131: If we want to have it working properly for GDPR and CCPA shouldn't we complicate a bit? 🙈 And change the line 131 into:
```js
return gdprConsent.gdprConsent === true && ccpaOptions.getSignal() === true;
```

src/tracking/CookiesBaker.js
```diff
@@ -0,0 +1,28 @@
+import Cookies from 'js-cookie';
+import { getCookieDomain } from '../shared/utils';
+import { COOKIES } from './cookie-config';
+
+export default class CookiesBaker {
```
nandy-andy, line 5: Nice name <3 :D

src/tracking/tracking-params-config.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+]
+
+function isPageView(event) {
+    const name = event.name.toLowerCase();
+    return name === 'view' || name == 'pageview';
+}
+
+function increment(value) {
+    return (parseInt(value) || 0) + 1;
```
nandy-andy, line 56: What do you think of adding the radix to `parseInt()` just in case? 😊

README.md
jakubbanas: I also wonder if closing it would break our CI, maybe we'd need some adjustment on our side as well.

src/tracking/CookiesBaker.js
```diff
@@ -0,0 +1,28 @@
...
+        let domain = getCookieDomain(window.location.hostname);
+        cookies.forEach(cookie => {
+            if (!cookie.options.hasOwnProperty('domain')) {
+                cookie.options.domain = domain;
+            }
+        })
+        this.cookies = cookies;
+    }
+
+    setOrExtendCookies(cookieValues, orgCookiesJar = {}) {
```
jakubbanas, line 16: Shouldn't it be `bakeCookies` then? 😄

src/tracking/DataWarehouseEventsSender.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    send(event) {
+        if (!event) return;
+
+        const path = getEventPath(event.name);
+        delete event.name;
+
+        const controller = new AbortController();
+        const urlParams = Object.keys(event)
```
jakubbanas, line 24: There is a handy standard library class to handle URL params, maybe it would be good to use it here to avoid diving into params implementations details.

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams
You could use the `append` method to add params, and `toString` to parse it into a string.

src/tracking/EventsTracker.js
```diff
@@ -0,0 +1,76 @@
...
+class TrackableEvent {
+    name;
+    platform;
+    env;
+
+    //onDWHTrackingComplete
+    //couldBeTrackedWithoutConsent
+}
+
+function invalidEvent(event) {
```
jakubbanas, line 15: Maybe `isInvalidEvent` would be more descriptive since it returns a boolean.

src/index.js
mfarul: Well, stillI do not understand fully the flow :(, ccpa is covered with the first part as it fires and sets gdprConsent = true.
This whole idea is working as GDPR and CCPA regions are not overlaping.

mfarul: i'm putting gdrpConsent === true && ccpaSignal === false

README.md
mfarul: I plan to move tracking code into another lib. Now i use track-opt-in just to prove that the whole idea will work.

PR #240, opened 2022-03-11
ADEN-11653 Adding Geo cookie if not exists

dmnsobczak:


src/shared/GeoManager.js
```diff
@@ -89,6 +90,56 @@ function getGeoDataFromCookie(type = 'country') {
...
+
+            request.send();
+        } catch (err) {
+            console.warn('no geo cookie found');
+            reject();
+        }
+    }).then((geoData) => {
+        Cookies.set('Geo', JSON.stringify(geoData), {
+            domain: getCookieDomain(window.location.hostname),
+            expires: 365,
```
mfarul, line 137: That's huge, f.e. Varnish set's this as a session cookie. Let's make more like 30 days?

src/shared/GeoManager.js
```diff
@@ -89,6 +90,56 @@ function getGeoDataFromCookie(type = 'country') {
     return false;
 }
 
+export function ensureGeoCookie() {
+    if (Cookies.get(COUNTRY_COOKIE_NAME)) {
+        return Promise.resolve();
+    }
+
+    const GEO_SERVICE_URL = 'https://services.fandom.com/geoip/location';
```
mfarul, line 98: will this work for dev env?

src/shared/GeoManager.js
dmnsobczak: In my browser I see that the cookie is set for one year 🤔

src/shared/GeoManager.js
dmnsobczak: Yes, they both gonna request same endpoint.

PR #249, opened 2022-06-13
ADEN-11905 Enable GPC signal in GDPR geo

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-11905

### DIFF
--- a/src/gdpr/ConsentManagementPlatform.js
+++ b/src/gdpr/ConsentManagementPlatform.js
@@ -98,7 +98,7 @@ class ConsentManagementPlatform {
             return false;
         } else if (!this.geoRequiresTrackingConsent()) {
             return true;
-        } else if (window.navigator.globalPrivacyControl) {
+        } else if (this.isGpcEnabled()) {
             return false;
         } else if (hasConsentCookie && this.optInManager.hasAcceptedTracking()) {
             return true;
@@ -166,7 +166,7 @@ class ConsentManagementPlatform {
                 this.checkUserConsent();
             });
 
-        if (window.navigator.globalPrivacyControl) {
+        if (this.isGpcEnabled()) {
             this.tracker.trackGpcImpression();
         }
     }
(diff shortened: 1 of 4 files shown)


src/gdpr/ConsentManagementPlatform.js
```diff
@@ -171,7 +173,12 @@ class ConsentManagementPlatform {
                 this.onAcceptTracking();
                 break;
             case false:
-                this.onRejectTracking();
+                if (window.navigator.globalPrivacyControl) {
+                    this.tracker.trackGpcImpression();
```
dmnsobczak, line 177: Discussed offline: better place for it:
https://github.com/Wikia/tracking-opt-in/blob/804f70d9203e232e7d9e2d595166013e6907a628/src/gdpr/ConsentManagementPlatform.js#L155

src/gdpr/ConsentManagementPlatform.js
```diff
@@ -96,6 +96,8 @@ class ConsentManagementPlatform {
 
         if (this.isOnWhiteListedPage()) {
             return false;
+        } else if (window.navigator.globalPrivacyControl) {
```
dmnsobczak, line 99:
```suggestion
        } else if (this.geoRequiresTrackingConsent() && window.navigator.globalPrivacyControl) {
```
And `this.geoRequiresTrackingConsent()` can be stored inside a new variable like `hasConsentCookie`

dmnsobczak (changes requested):
Enable only in GDPR countries, not worldwide

src/gdpr/ConsentManagementPlatform.js
```diff
@@ -163,6 +165,10 @@ class ConsentManagementPlatform {
 
                 this.checkUserConsent();
             });
+
+        if (window.navigator.globalPrivacyControl) {
```
dmnsobczak, line 169: Some old browsers might not implement `window.navigator` 🤔 
https://github.com/Wikia/tracking-opt-in/blob/92447e68b13d79330e1b74dde74105808a17db6c/src/shared/LangManager.js#L8

Can we move it to the separate method `isGpcEnabled() { return window.navigator && window.navigator.globalPrivacyControl; }` and call it in 3 places in this file?

PR #252, opened 2022-07-04
ADEN-11989 Include uppercase letters in the random string for beacon

nandy-andy:


src/shared/CookieManager.js
```diff
@@ -14,17 +14,22 @@ class CookieManager {
...
         return resultValue;
     }
 
-    generateValue(name, withTimestamp) {
+    generateValue(withTimestamp) {
         // for users from GDPR countries if they did not give consent for tracking
         // we assign random values to session cookies; should match results of:
         // https://developer.fastly.com/reference/vcl/functions/randomness/randomstr/
-        let resultValue = (Math.random().toString(36) + '_________________').slice(2, 12)
+        const charsToPickFrom = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXTZabcdefghiklmnopqrstuvwxyz_';
```
preludek, line 27: `validCharacters` maybe?

src/shared/CookieManager.js
```diff
@@ -14,17 +14,22 @@ class CookieManager {
...
+
+        let resultValue = '';
+        for(resultValue; resultValue.length < 10;) {
```
preludek, line 30: We dont need resultValue here - `for(; resultValue.length < 10;) {` would work the same. I'd prefer while loop here personally 🤷

src/shared/CookieManager.js
```diff
@@ -14,17 +14,22 @@ class CookieManager {
...
+            resultValue += charsToPickFrom[(Math.random() * charsToPickFrom.length) | 0];
```
preludek, line 31: I think I don't understand that `|  0` part 🙈. Why dont we just floor the random * charsToPickFrom.length?

preludek (approved):
Some nit picks

src/shared/CookieManager.js
nandy-andy: We need integers as the array's key are integers and `| 0` is shorter to write than `Math.floor()` 😈

PR #257, opened 2022-10-12
ADEN-12305 add new partners

jakubbanas:


src/shared/consts.js
```diff
@@ -181,6 +181,8 @@ export const IAB_VENDORS = [
     281, // Wizaly
     32,  // Xandr, Inc. (AppNexus Inc.)
     173, // YieldMo
+    2,   // ATP: Captify Technologies Limited
```
dmnsobczak, line 184: Partners are in alphabetical order by name, for easier browsing :)

dmnsobczak (changes requested):
Alphabetical reorder request

src/shared/consts.js
jakubbanas: Done 👍

dmnsobczak (approved):

PR #262, opened 2022-11-30
CATS-3457 | Reduce the bundle size

mszabo-wikia:
As part of our mobile editor performance investigations, we noticed that the tracking-opt-in library was a rather large asset, impacting mobile editor load times on slower connections. Attempt to reduce the impact by doing a quick optimization pass:

* Don't load preact twice - use the ESM version consistently.
* Minify CSS output using `cssnano`.

This cuts down the bundle size from ~273 kB (~73 kB gzipped) to ~234 kB (~67 kB gzipped).

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
         "consent-string": "^1.2.4",
         "cross-env": "^5.1.4",
         "css-loader": "^0.28.11",
+        "cssnano": "^4.1.11",
         "es6-promise": "^4.2.4",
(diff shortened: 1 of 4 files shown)


jakubbanas (approved):

kasperekt (approved):

nandy-andy (approved):

dmnsobczak (approved):

dmnsobczak:
Tested, works fine 👍 Merging 🚀

PR #265, opened 2022-12-02
ADEN-12407 - OneTrust improvements

domjanik:


src/onetrust/index.js
```diff
@@ -37,7 +37,7 @@ class OneTrustWrapper {
     }
 
     setTrackingOptInCookies() {
-        if (getCookieValue('OptanonConsent').indexOf(this.ALLOW_TRACKING_GROUP) !== -1) {
+        if (window.OptanonActiveGroups.indexOf(this.ALLOW_TRACKING_GROUP) !== -1) {
```
jakubbanas, line 40:
```suggestion
        if (window.OptanonActiveGroups?.indexOf(this.ALLOW_TRACKING_GROUP) !== -1) {
```
In case the window variable is not set.

Also, I recommend [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method which is easier to read.

nandy-andy (approved):

PR #269, opened 2022-12-28
DE-5149 | Fix pv_number counting

JackThePie:
## Ticket
https://fandom.atlassian.net/browse/DE-5149

## Description
Change pv_number initialization to work with unified-pltaform changes: https://github.com/Wikia/unified-platform/pull/12256

### DIFF
--- a/src/gdpr/ConsentManagementPlatform.js
+++ b/src/gdpr/ConsentManagementPlatform.js
@@ -227,8 +227,14 @@ class ConsentManagementPlatform {
         const domain = getDomain(window.location.host);
         const path = '/';
 
+        const pathSplitted = window.location.pathname.split('/');
+        const languagePath = pathSplitted[1] !== 'wiki' ? pathSplitted[1] : 'en';
+
         Cookies.set('tracking_session_id', sessionId, {domain, expires, path});
-        Cookies.set('pv_number', pvNumber + 1, {expires, path});
+        Cookies.set('pv_number', {
+            ...pvNumber,
+            [languagePath]: pvNumber[languagePath] + 1
+        }, {expires, path});
         Cookies.set('pv_number_global', pvNumberGlobal + 1, {domain, expires, path});
     }
 


src/gdpr/ConsentManagementPlatform.js
```diff
@@ -227,8 +227,14 @@ class ConsentManagementPlatform {
         const domain = getDomain(window.location.host);
         const path = '/';
 
+        const pathSplitted = window.location.pathname.split('/');
+        const languagePath = pathSplitted[1] !== 'wiki' ? pathSplitted[1] : 'en';
```
nandy-andy, line 231: Consider extracting lines 230 and 231 to a separate method and unit test it :)

nandy-andy: Although, it doesn't work for Wiedźmin wiki which is in `pl`: https://wiedzmin.fandom.com/wiki/Wied%C5%BAmin_Wiki But maybe "content language" is not the same as the language path. But then the same `languagePath` will be set for visits on: https://wiedzmin.fandom.com/wiki/Wied%C5%BAmin_Wiki and https://witcher.fandom.com/wiki/Witcher_Wiki.

src/gdpr/ConsentManagementPlatform.js
```diff
@@ -227,8 +227,14 @@ class ConsentManagementPlatform {
...
+
         Cookies.set('tracking_session_id', sessionId, {domain, expires, path});
-        Cookies.set('pv_number', pvNumber + 1, {expires, path});
+        Cookies.set('pv_number', {
```
nandy-andy, line 234: You probably need to cover also the previous version when the cookie is set and it's a number instead of an object. With this implementation when a cookie is set it'll get overwritten by: 
```
{
  [languagePath]: NaN
}

src/gdpr/ConsentManagementPlatform.js
nandy-andy: Oh, I can see now it's supposed to be the content language: https://github.com/Wikia/unified-platform/pull/12256/files#diff-5e99c107d960a8089f6525054c9789505b43d1e0f353bb6c462e05ec12d55376R69 I think you'll need to rethink the solution then because in the given example we'll have an error again. The current issue will be gone as the PVs per language will be counted but the English PV number for given two communities will be incorrect in case when a user browse through both of the communities.

PR #282, opened 2023-05-24
ADEN-13227 Hide Reject All feature behind the flag

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/ADEN-13227

### Description
Hide Reject All feature behind the flag. Default option which will be released global should be set to false.

### DIFF
--- a/src/shared/GeoManager.js
+++ b/src/shared/GeoManager.js
@@ -73,6 +73,7 @@ const COUNTRIES_REQUIRING_PROMPT = [
     'uk', // United Kingdom of Great Britain and Northern Ireland
     'wf', // Wallis-et-Futuna
 ];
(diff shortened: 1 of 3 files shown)


src/shared/GeoManager.js
```diff
@@ -156,6 +157,10 @@ class GeoManager {
     hasGeoCookie() {
         return this.country !== MISSING_COOKIE_NAME;
     }
+
+    hasRejectAllFunctionality() {
+        return COUNTRIES_WITH_REJECT_ALL_FUNCTIONALITY.indexOf(this.country) !== -1;
```
domjanik, line 162: You could also use includes here :) 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

szymon-nawrocki:
Reopening

szymon-nawrocki:
Reopening

PR #314, opened 2023-12-13
TACO-269 - Added blocked purposes filtering

domjanik:
In this PR I've added Legitimate Interests filtering. Since TCF 2.2 some Legitimate Interests should be blocked, and TCF strings, containing these interests are being recognized as invalid by Pubmatic. I've modified our logic to get only these interests, that are not blocked.


src/gdpr/ConsentManagementProvider.js
```diff
@@ -22,6 +22,7 @@ const PROVIDER_CONSENT_COOKIE_NAME = 'addtl_consent';
 const VENDOR_LIST_URL_BASE = 'https://script.wikia.nocookie.net/fandom-ae-assets/tcf/v2.2/';
 const VENDOR_LIST_FILE_NAME = 'vendor-list.json';
 const VENDOR_LIST_VERSION_NAME = 'archives/vendor-list-v[VERSION].json';
+const BLOCKED_INTERESTS = [1,3,4,5,6];
```
jbienkowski-fandom, line 25: I guess we should adapt this part of code as well: https://github.com/Wikia/identity-engine/blob/main/src/services/OptInService/consentProviders/tcf-api.ts#L13

jbienkowski-fandom (approved):

PR #320, opened 2024-01-25
TACO-390 Update privacy string for kid users

szymon-nawrocki:
### Ticket
https://fandom.atlassian.net/browse/TACO-390

### Description
Currently in US when user clicks "I am a kid" in Age Gate modal, the `usprivacy` cookie stays at the value `1YNN` which means - NO opt out.

In this PR I fixed this issue, so when user is on wiki directed at children, and he clicks "I am a kid", the `usprivacy`cookie is updated to `1YYN` which means opt out.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,7 @@ import UserSignalMechanism from './ccpa/UserSignalMechanism';
 import { communicationService } from './shared/communication';
 import { debug } from './shared/utils';
 import { oneTrust } from './onetrust';
+import Cookies from 'js-cookie';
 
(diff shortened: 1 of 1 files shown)


domjanik (approved):

domjanik:
LGTM 👍 We've to remember about this, during OneTrust migration. This functionality should be maintained after that migration.

szymon-nawrocki:
The issue was fixed on the `identity-engine` side in this PR:
https://github.com/Wikia/identity-engine/pull/254

Closing.

PR #330, opened 2024-03-25
TACO-503 Send pv_unique_id to gdpr_events

szymon-nawrocki:
### Description
Send `pv_unique_id` to `gdpr_events` data table


src/gdpr/Tracker.js
```diff
@@ -22,6 +23,10 @@ function getBeaconFromCookie(cookieName) {
     return Cookies.get(cookieName || DEFAULT_BEACON_COOKIE_NAME);
 }
 
+function getPvUniqueId() {
+    return window.fandomContext?.tracking?.pvUID
```
domjanik, line 27: I'm wondering if it would be possible to get pvUID for impression event? AFAIR - pvUID in fandomContext is being set in IE after we get the consent from user. Maybe we could do a fallback to mw variable(MW) and window.pvUID(F2) ?

domjanik (changes requested):

src/gdpr/Tracker.js
szymon-nawrocki: It will be possible with this code, because this Tracker is waiting for event `'identity ready'` from identity-engine. (line 53)

The event will be sent once identity is ready, so both `beacon` and `pv_unique_id` will be available then :)

domjanik (approved):

